PR #18, opened 2015-01-28
Simplify the config structure to improve usability

lewisnyman:
I'm giving a presentation on this tool next week, comparing how simple it is to use compared to a tool like PhantomCSS. I've been including config examples in my slides which means I've been thinking a lot about how to simplify them so the on boarding process on getting up and running is even smoother.

  {
     "name": "phone",
     "viewport":  {"width": 320, "height": 480 }
  }

Why wrap the dimensions in another object? Each viewport is already wrapped in the array 'viewports' so it feels like we could simplify the structure here.

  {
     "name": "phone",
     "width": 320, 
    "height": 480 
  }

  "grabConfigs" 

The name of this array is a little nondescript, it would be better if it described the items it contains like the "viewports" array does. How about "pages" or "tests"?

  {
    "testName":"Home",
    "url":"http://example.com/"
  }

If we rename the "genConfig" array to be more descriptive, we could probably just rename "testName" to "name".

This will break existing builds so it would have to be included in a major version update or we just support both the new and existing config structure for a while.

What do you think?

### DIFF
--- a/capture/config.default.json
+++ b/capture/config.default.json
@@ -2,20 +2,23 @@
 	"viewports" : [
 		{
 		 "name": "phone",
-		 "viewport": {"width": 320, "height": 480}
+		 "width": 320,
+     "height": 480
 		}
 		,{
 		 "name": "tablet_v",
-		 "viewport": {"width": 568, "height": 1024}
+		 "width": 568,
+     "height": 1024
 		}
 		,{
 		 "name": "tablet_h",
-		 "viewport": {"width": 1024, "height": 768}
+		 "width": 1024,
+     "height": 768
 		}
 	]
-	,"grabConfigs" : [
+	,"scenarios" : [
 		{
-			"testName":"http://getbootstrap.com"
+			"label":"http://getbootstrap.com"
 			,"url":"http://getbootstrap.com"
 			,"hideSelectors": [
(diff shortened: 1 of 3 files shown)


garris:
Hi @lewisnyman -- I pulled this into master but don't have time to test today.  Please take a look if you get a chance.   In particular, try `gulp echo` also -- these changes may have broken that.  (There is probably a better way to manage that feature but for now it is a clunky extra file.)

garris:
Just a followup.  The changes here check out on my end.   There was an issue with `gulp echo` but I have fixed that and pushed it to master.

capture/config.default.json
```diff
@@ -2,20 +2,23 @@
 	"viewports" : [
 		{
 		 "name": "phone",
-		 "viewport": {"width": 320, "height": 480}
+		 "width": 320,
+     "height": 480
```
jehoshua02: Inconsistent indentation. I don't know about you guys, but I use 2 spaces instead of a tab for all javascript. Configured my editor to do it that way when I hit the tab key, and to auto indent on new line.

garris: @jehoshua02     Please go ahead and format this file.   I will pull your changes.

Issue #23, opened 2015-01-28
CasperJS and PhantomJS are BackstopJS Dependencies.


jehoshua02:
CasperJS and PhantomJS are BackstopJS dependencies. Please consider adding them to `package.json` and/or `bower.json`.

jehoshua02:
Maybe looking at how I've done things here will help: https://github.com/jehoshua02/backstopjs-example

garris:
This is a very popular enhancement.  Do you care to make a pull request?

Also, is it possible to implement this install without breaking the old install option.  There may be users who don't necessarily want their existing phantomjs & casperjs versions bumped.

What do you think?

jehoshua02:
You would bump your version number and they could stay on an older version of backstop.

jehoshua02:
I've forked it and will submit a pull request sometime.

garris:
Done.  Closed with release https://github.com/garris/BackstopJS/releases/tag/1.2.0

PR #25, opened 2015-01-28
Travis CI config

lewisnyman:
This is some very basic Travis CI config that makes sure the code runs without any hard errors.

Not bullet proof but it's something to build on.

You can see the builds from my repo here: https://travis-ci.org/lewisnyman/BackstopJS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: node_js
 node_js:
   - "0.10"
 install:
-  - npm install gulp
+  - npm install gulp -g
+  - npm install phantomjs -g
+  - npm install casperjs -g
   - npm install
-script: gulp test
+script: gulp reference
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: node_js
+node_js:
+  - "0.10"
+install:
+  - npm install gulp
+  - npm install
+script: gulp test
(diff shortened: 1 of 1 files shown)


garris:
Hi Lewis, this is great work.   I see on Line 273 & line 309 of the Travis log that there is an error reported by BackstopJS.  That error is thrown when backstopjs first runs and finds missing dependencies (this is after backstopjs attempts to spawn a bower process to install dependencies for the reporting module).   I think this is easily fixable by simply adding those dependencies to the build repo instead of omitting them like they are now.   

In any case this Travis stuff is useful and I will merge that in.   Would love to clean up this install.   @jehoshua02 suggested some nice improvements in #23 -- How do those look to you?

lewisnyman:
Yes, it would be great if the only command users had to run was npm install. I'd rather stick with the Travis config that maintains the status quo, and then make changes that affect everyone in dedicated issues. How does that sound?

garris:
Yes.  Totally makes sense.

lewisnyman:
The only thing left to do here is for you to do here is to head to https://travis-ci.org and login, then enable your repository for testing.

You can also add status images to the Readme so it's easy to tell if the latest changes broke the script http://docs.travis-ci.com/user/status-images/

garris:
Done and done!
screen shot 2015-01-31 at 10 04 13 am

Issue #26, opened 2015-01-30
Install BackstopJS via NPM


jehoshua02:
The benefit of installing BackstopJS via NPM is that it's dependencies would be installed recursively.

```shell
npm install --save-dev garris/BackstopJS
```

This not only records BackstopJS as a dependency in my project's `package.json` and installs it into `node_modules/`, but it also installs BackstopJS dependencies, recorded in `node_modules/BackstopJS/package.json`, into `node_modules/BackstopJS/node_modules`.

I'm not sure I would even keep the `bower.json` at that point. If a user has `bower` they have `npm`.

jehoshua02:
See [backstopjs-example](https://github.com/jehoshua02/backstopjs-example). Also see related, but independent issue [CasperJS and PhantomJS are BackstopJS Dependencies](https://github.com/garris/BackstopJS/issues/23).

garris:
Yes, let's do this.  We'll need to remember to include the compare/bower_components/...  assets when creating the NPM entry.  I will need to update various documentation as well.

garris:
ok.  I have created an NPM branch with necessary bower dependencies already added in github and I have also instantiated npmjs.com/package/backstopjs.

Let me know if you are ok for creating a pull request with the changes you're proposing.   I think it's best to simply start with a moving to an npm-only install.  After making sure that is solid we can do another release where we add hooks to installing phantom & casper etc.

Just let me know if you'll have time/interest to create the pull request.  

I very much appreciate the help!

Thanks!

jehoshua02:
@garris I'm excited to see activity in this project. I may or may not have time to submit a pull request this weekend. I will try but I don't want to hold back progress if I can't find the time.

garris:
No problem at all!  I really appreciate your interest in the project!   

You can post here if you decide to pick up the task to let anyone else who is working know. 

Best regards!

jehoshua02:
@garris For npm install, is there anything else to do other than [updating the README.md](https://github.com/garris/BackstopJS/pull/27)?

PR #27, opened 2015-01-31
Update README.md for NPM Install

jehoshua02:
+ Changes all usage over bower over to npm.
+ Changes references to `./bower_components` to `./node_modules`.


jehoshua02:
Related issue: https://github.com/garris/BackstopJS/issues/26

garris:
@jehoshua02 There is [an article here](http://css-tricks.com/automating-css-regression-testing/) which we should really update prior to releasing this change.

jehoshua02:
@garris I'll let you handle that one. But there's nothing else in the repository?

garris:
I don't think so.   Great progress!

PR #28, opened 2015-02-01
Split Gulp Tasks into Separate Files

jehoshua02:
Uses `require-dir` to require all task files in `gulp/tasks` directory.

Carefully split out each task one by one and make sure no requires or variables were missing.

Leftover code after splitting out all tasks is saved in `_init_.js` and should still run 
whenever gulp is run since `require-dir` requires all files it can find.

This was a pure refactor that should not have affected functionality.


jehoshua02:
@garris I'm on it. Don't freak out. I'll fix it :)

jehoshua02:
@garris That's better. All done. Merry Christmas.

garris:
Boom. Please verify!

PR #29, opened 2015-02-01
Format Default Configs

jehoshua02:
Used "Fomat: Javascript" command, made sure objects in array started and
ended on their own line and were indented one more.


garris:
Much Gratitue.

PR #37, opened 2015-02-16
Add title

ghiden:
Thank you for creating this awesome tool for visual testing.
I noticed that the home page for BackstopJS does not have a title.


garris:
Wow, nice catch -- will take that straight away!

Issue #38, opened 2015-02-22
Empty markup returned when testing HTTPS pages.


Chapabu:
If your configuration is set up to hit a page over HTTPS, then you get the following markup:

``` bash
CasperJS: CREATING NEW REFERENCE FILES
CasperJS: LOG> CASPER IS RUNNING
CasperJS: LOG> PHANTOM IS RUNNING
CasperJS: 
==================
Current location is https://www.twitter.com
==================

CasperJS: 

<html><head></head><body></body></html>
CasperJS: 
======================
echoFiles has completed 
=======================


Echo files completed.
```

This isn't an issue with BackstopJS exactly, but how PhantomJS handles pages running under HTTPS. Following advice on StackOverflow [here](https://stackoverflow.com/questions/26415188/casperjs-phantomjs-doesnt-load-https-page/26419531#26419531), it can be fixed by passing `--ssl-protocol=tlsv1` in with the `tests` array prior to spawning the CasperJS process (e.g. [echo.js](https://github.com/garris/BackstopJS/blob/master/gulp/tasks/echo.js#L8-L13)).

Obviously this doesn't make sense to add this in globally, but perhaps we could add an option for which SSL protocol to use (if any) in the `backstop.json` config file? I guess it'd be most suitable in the scenarios section?

I.e.

``` json
...
 "scenarios": [
    {
      "label": "Twitter",
      "url": "https://www.twitter.com",
      "ssl-protocol": "tlsv1",
      "hideSelectors": []
    }
 ]

```

I'm happy to work on a pull request for this, but does the suggested approach make sense?

garris:
Hi @Chapabu  thank you for filing this and researching the issue!   This suggestion makes sense.  Question tho --  is there any case where we'd need to set ssl-protocol if the url didn't contian `"https://"`?  My guess is no -- and if that guess is correct then my instinct is to handle this internally with a switch on `if (/^https:\/\//.test(url)) {...}`  just prior to the spawn block.   Maybe this is a nice convienience for nearly every (if not all) use cases?   What do you think?

Chapabu:
[...]

PR #39, opened 2015-02-23
Add HTTPS protocol option to CasperJS spawn task.

Chapabu:
Added args array for holding arguments to pass to CasperJS/PhantomJS.
Concatenated tests and args array into a casperArgs array to pass to spawn(casperProcess).

### DIFF
--- a/gulp/tasks/echo.js
+++ b/gulp/tasks/echo.js
@@ -7,10 +7,14 @@ gulp.task('echo',function(){
 
   var tests = ['capture/echoFiles.js'];
 
+  var args = ['--ssl-protocol=any'];
+
+  var casperArgs = tests.concat(args);
+
   // var args = ['test'].concat(tests); //this is required if using casperjs test option
 
   var casperProcess = (process.platform === "win32" ? "casperjs.cmd" : "casperjs");
-  var casperChild = spawn(casperProcess, tests); //use args here to add test option to casperjs execute stmt
+  var casperChild = spawn(casperProcess, casperArgs); //use args here to add test option to casperjs execute stmt
 
   casperChild.stdout.on('data', function (data) {
     console.log('CasperJS:', data.toString().slice(0, -1)); // Remove \n
--- a/gulp/tasks/test.js
+++ b/gulp/tasks/test.js
@@ -45,11 +45,15 @@ gulp.task('test',['init'], function () {
 
   var tests = ['capture/genBitmaps.js'];
 
+  var args = ['--ssl-protocol=any'];
+
+  var casperArgs = tests.concat(args);
+
   // var args = ['test'].concat(tests); //this is required if using casperjs test option
 
   // var casperChild = spawn('casperjs', tests);//use args here to add test option to casperjs execute stmt
   var casperProcess = (process.platform === "win32" ? "casperjs.cmd" : "casperjs");
-  var casperChild = spawn(casperProcess, tests);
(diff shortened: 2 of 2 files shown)


Chapabu:
This should fix #38. I've added another array so as not to confuse the `tests` array. I've then concatenated this array with the previous tests array.

Chapabu:
See #38

garris:
Super!  Thank you for the pull request.  Will pull later today.  Best GS

garris:
Looks good!    @Chapabu do you have a backstop.json config you used to test this?

Chapabu:
Sure, I've added a gist here: https://gist.github.com/Chapabu/787ddfebd5cf6d751278

I tested against 3 HTTPS pages and 3 non-HTTPS pages to ensure nothing was broken.

garris:
So, here is something interesting.   I used your test file on the previous backstopjs version 0.4.1 and it seems to have rendered everything fine.  See attached.   

Maybe this issue only applies to certain versions of Phantom?   

What version of Phantom are you running?  And what OS?

```
garriss-macbook:backstopjs gshipon$ phantomjs -v
1.9.8
```

screen shot 2015-02-23 at 3 46 23 pm

Chapabu:
I'm running the following:

OSX Yosemite
PhantomJS 1.9.7
CasperJS 1.1.0-beta3

Although I have just noticed that PhantomJS 1.9.8 has changed the default ssl protocol to TSLv1 instead of SSLv3 ([release notes](http://phantomjs.org/release-1.9.html)), so this looks like the main problem was me being on the previous release of PhantomJS. That makes me feel slightly silly as it's the first thing I should have checked!

So yeah, this issue is/was only apparent on PhantomJS < 1.9.8.

garris:
Cool.  Mystery solved!    I am still tempted to keep the change in however -- I don't mind being a little bit more compatible with older versions.   My only concern would be that adding this flag could somehow break the current or future versions of Phantom. 

Or, I can just as easily back out the change.

Any thoughts there?

Chapabu:
I think it's useful for now, as it allows Backstop to work for HTTPS pages in all _available_ versions of PhantomJS. However, I see the concern with potential future breakage. I guess the only real failsafe would be to tweak the `.travis.yml` file and add [matrix variables](http://docs.travis-ci.com/user/environment-variables/#Matrix-Variables) for different versions. So you could explicitly run travis against 1.9.7, 1.9.8, and whichever is the latest version from npm. 

I think it'd probably look a little like this:

``` yaml
language: node_js
node_js:
  - "0.10"
env:
 - PHANTOM_VERSION='1.9.7'
 - PHANTOM_VERSION='1.9.8'
 - PHANTOM_VERSION='latest'
install:
  - npm install gulp -g
  - npm install phantomjs@PHANTOMJS_VERSION -g
  - npm install casperjs -g
  - npm install
script:
  - gulp reference
  - gulp test
```

That way travis would run against the older, `SSLv3` default argument, the newer `TSLv1` default argument, as well as checking it still works against the latest PhantomJS from NPM.

I can try a pull request for this as well if you'd like, but it'd have to wait until tomorrow as it's 12:30am where I am now :sleeping:

garris:
Thanks so much for your input on this thing.   I think that if that flag does not break Phantom 1.9.8 then we can just go forward with that.   I don't think we need to add the travis complexity.   If you get around to verifying with phantom 1.9.8 then that is great.   If not, I will try to get to it in a few days.   Thanks again!

PR #43, opened 2015-03-06
Created demo for new backstopjs users

Shashikant86:


garris:
Hi @Shashikant86,  Thank you very much for your pull request!   Couple of thoughts here...

1) The version of BackstopJS you're using is out of date.   Since that version we moved to a much simpler NPM install version which installs everything in one go.  The bower install version is now deprecated. 

2) I like what you did with the gulp-chug module -- but I am concerned that this would conflict with any preexisting gulp configs in a developer's working project directory.   I think it could be disruptive to developers to require this file at the root of their projects.  Also, gulp-chug seems to be blacklisted https://github.com/gulpjs/plugins/issues/93 -- On first glance it seems due to some philosophical differences.

Shashikant86:
@garris Thanks for your feedback. I have removed bower and gulp_chug dependency also updated README file accordingly. Let me know what you think ?

garris:
Thanks for your effort, but at this point I am having trouble making sense of this Demo at a high level.  I don't think this adds much value since BackstopJS documentation already walks users through set up.

Issue #52, opened 2015-04-02
Expose mismatch threshold as a configurable option


lewisnyman:
Right now it's possible for subtle regressions to be missed by the reporting tool. This is a shame because the great value in a tool like this is catching regressions that are easily missed by humans.

The default value right now is:
```
var misMatchThresh = 1;
```
See #21

garris:
I know, this is totally worth doing.  I have some time coming up where I could get it done. Will do a pass ping you when I have something to test.   GS

garris:
Files merged and bugs fixed.

MatteoWebDesigner:
Hi I guess something has changed even if I set mismatch 0.1, the value is not used. Is there an example showing the correct way to config this setting. Thanks

garris:
What version are you on?

MatteoWebDesigner:
1.1.0
config: https://github.com/MatteoWebDesigner/CSS-Architecture/blob/master/backstop.json

garris:
@MatteoWebDesigner 

this line 
https://github.com/MatteoWebDesigner/CSS-Architecture/blob/master/backstop.json#L54

should be

     "misMatchThreshold": 0.1,

(you have `"misMatchThresh": 0.1,`)

MatteoWebDesigner:
I see :O https://github.com/lewisnyman/BackstopJS/blob/f6d2a09d8556294f85fce1ee1bfa625143ce5db0/capture/echoFiles.js#L19

thank you now works fine.

garris:
Cheers.  :+1:

PR #53, opened 2015-04-03
Initial implementation of threshold configuration, with fallback to 1.

lewisnyman:
I don't think this is the best way to do it, but it seems to work. I've kept a default so the tests don't fail if the variable doesn't exist.

#52

### DIFF
--- a/capture/config.default.json
+++ b/capture/config.default.json
@@ -1,4 +1,5 @@
 {
+  "misMatchThreshold" : 1,
   "viewports": [
     {
       "name": "phone",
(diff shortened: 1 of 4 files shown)


garris:
I think it's pretty good.  The only thing I would consider is moving that parameter to the scenario node so that with a selector like `body` one could use a higher number and with a selector like `footer` one could use a lower value.  I think these are relevant cases.  What do you think?

garris:
Hey Lewis -- awesome work!   There was one funny namespace bug in there which I fixed.   I didn't have a chance to test the final -- will do tonight or tomorrow AM.   Just wanted to thank you for the update!   

You can install from here to test...
    npm install garris/backstopjs#master

lewisnyman:
Nice! Thanks :) I got a little stuck when working on it

PR #69, opened 2015-06-19
Split report html

YuliaTsareva:
I splitted compare/index.html into separate html, js and css files in order to simplify work with them.


garris:
Merged.

PR #75, opened 2015-07-07
WordPress has a capital P

aaronjorbin:
Pretty minor, but it's WordPress, not Wordpress.


garris:
@aaronjorbin  Nice attention to detail!  Thanks.  GS

Issue #90, opened 2015-09-10
Add an explicit gulp task to compare images


klausbayrhammer:
We are currently trying to integrate backstop tests into our CI/CD pipeline but the problem is that I currently don't see a way to pass the results of a backstop test to the gulp task (or the caller).

Is it correct that the comparison of reference and test images in done on the fly in `compare.js#compareTestPair`? What we'd require is a gulp task (or some kind of CLI) to explicitly trigger the image comparison and retrieve the result as a return value. Our overall goal is to run the backstop tests on our build servers and stop our build when errors occured.

Probably this is related to #32 and maybe we could use [node-resemble-js](https://www.npmjs.com/package/node-resemble-js) to compare the images.

klausbayrhammer:
The following snippet could probably solve the problem. @garris if you think this is going in the right direction I'd clean the things up and provide a PR.

```javascript
const gulp = require('gulp');
const resemble = require('node-resemble-js');
const _ = require('underscore');

gulp.task('compare_backstop_images', function () {
    var compareData = getCorrectRelativePath(require('./backstop.json').paths.compare_data);
    var testPairs = require(compareData).testPairs;

    _.each(testPairs, function (pair) {
        console.log('testing', pair.fileName);
        var referencePath = getCorrectRelativePath(pair.reference);
        var testPath = getCorrectRelativePath(pair.test);

        resemble(referencePath).compareTo(testPath).onComplete(function(data) {
            if (data.misMatchPercentage > pair.misMatchThreshold) {
                console.log(data);
                console.log('Mismatch exceeded threshold for image', pair.fileName);
                process.exit(1);
            }
        });
    });
...
```

garris:
@klausbayrhammer this looks like a great solution/feature.  Would this behavior be compatible with the SlimerJS engine option?  Would this be an option that runs in addition to / independent of the in-browser reporting process?

klausbayrhammer:
I'm not sure if this works in all environments but we are running the backstop tests/image comparisons in a centos container with slimer installed and it works for us.

For now I'd rather provide this task as an independent addition, so we don't have any trouble when it comes to backwards compability.

[...]

PR #93, opened 2015-09-15
Add an explicit gulp task to compare images #90

klausbayrhammer:


garris:
@klausbayrhammer, I merged the PR added config flags and cleaned up the workflow.  I think this is a great feature.  Thank you for your help!

Issue #102, opened 2015-10-15
Feature: Ability to run casperjs code linked from a scenario


sanchothefat:
I can have a stab at this but Casperjs / Phantomjs are pretty new to me.

My thinking is that where we define a scenario you could reference a casperjs file to run when the readyEvent / delay have happened eg:

```json
{
  "viewports": {
    ...
  },
  "scenarios": [
    {
      "label": "http://example.com",
      "url": "http://example.com",
      "onReadyEvent": "./path/to/casperjs/file.js"
    }
  ]
}
```

Offers full flexibility for those who want to go a bit deeper. I'm looking into the `test.js` to see where / if this could work but any thoughts in the meantime on the idea would be great to hear.

garris:
Hi @sanchothefat.  Thanks for looking into this.  This is a feature that has been requested before.   Here are some comments/ideas...

- maybe change `onReadyEvent` to `onReadyScript` -- just to make it clear that we are linking to a file.

- the ready event handler you are looking for is here...
https://github.com/garris/BackstopJS/blob/master/capture/genBitmaps.js#L97

- in that spot you would probably want to do something like this...
```
    if (scenario.onReadyScript && fs.isFile(scenario.onReadyScript)) {
      var executableJsAsString = fs.read(scenario.onReadyScript);
      // execute that code here
      // or you may be able to import the file in some other way -- just a suggestion
    }
```

- please contemplate this feature and how the two features might interact or support each other...
https://github.com/garris/BackstopJS/issues/94

Let me know if I can help.

Cheers.

sanchothefat:
Awesome, thanks @garris - this is a great place to start. Agreed on your first point, the naming is much clearer, I momentarily forgot it was a json file, no functions :)

#94 looks interesting too. Perhaps a prepackaged script and `scenario.onReadyScript` can be an array. Something along those lines. I'll have a play.

garris:
@sanchothefat this is nice work!  But I am a little concerned that an array of scripts is overcomplicated.  I think the use case is likely to be more simple for most users.   For the minority of users who need multiple file support they can always concat those things in gulp or grunt exactly how they like.   This would reduce complexity on the BackstopJS side -- reading files and handling errors will be an unnecessary pain...

In relation to https://github.com/garris/BackstopJS/issues/94   -- I only mention this because I imagine it makes sense that an `onReadyScript` would run on each of those sub-scenarios -- and that may or may not be a good thing.

[...]

Issue #94, opened 2015-09-22
feature request: selctor-based link crawling


garris:
A `"crawlSelectors": []` property will be added to each scenario in the backstop.json config file.  For each matching crawlSelector, a collection of matching elements will be returned.  For each matching element the original URL will be reloaded and the current element index will receive a click event. The resulting page or state will be loaded and tested. 

Selectors are followed to a depth of 1. The child scenario will inherit all the test properties from the parent scenario. This would enable rudimentary testing of various page links, pop-up and control states.

parzerphilipp:
are there any new updates on this topic?
or should https://github.com/fffunction/backstop-crawl be used?

garris:
This never happened.   Probably best done with external solution.

PR #103, opened 2015-10-15
On ready script for scenarios - feature #102

sanchothefat:
Hi @garris - this a first pass at #102. Works pretty well so far, I just have a simple test where I click a menu toggle in a custom file like so:

```js
module.exports = function( casper, scenario ) {
	casper.click( '.menu-toggle' );
};
```

It'll take an array of scripts as I thought it'd be handy to have a common library eg. for logging in etc... There's also a config option for setting the base path to the scripts so they can be included shorthand.

While I'm not 100% certain on the way to spawn a child scenario re #94 I've written it so that the array of scripts can have custom built in casperjs scripts added to it early in the scenarios loop.

Let me know what you think so far. Not sure I needed to modify the `paths.js` file to be honest, I haven't really tested without a custom config file yet.


garris:
@sanchothefat  Beautiful!  Very well done -- thank you.  I haven't had a chance to run the code yet but it looks solid.  I will make some minor tweaks and feature this in the next release. It could take a week or so.  Have you attempted to use this for log-in anywhere?  I think that will be a popular use-case.

sanchothefat:
@garris cool, yeah I thought you'd want to make some version number tweaks etc. I haven't tried using it to log in yet but having it running against a local WP site. I'll try it out.

jkopcsek:
Hi! 
Thanks dudes for this great and simple regression testing suite. I really like this new feature, but i wonder: is it possible to wait for some time before returning and making the screenshot? There's no callback in the onReadyScript-function, so casper.wait isn't very useful. Or am I missing something?

sanchothefat:
@jkopcsek in my tests `casper.wait()` does indeed work within the external casper script so you should be able to pretty much any additional casper code you want before it continues to the screenshot stage.

jkopcsek:
Thanks a lot. Looking forward to build some more sophisticated tests!

PR #104, opened 2015-10-16
Adding label to console log compare output.

bradwade:
This adds the label to the comparison results shown in command line report.

Currently the results are too cryptic to decipher as the filename doesn't contain information to let the user know which page/component is failing/succeeding. (A unique number is used at the beginning of the filename, but that number doesn't intuitively translate into a specific page/component to the user running the test.)


garris:
@bradwade Thanks for the PR!   Will include in next release.  Cheers.

PR #107, opened 2015-10-26
Add Casper CLI flags support

### LINKED ISSUE #92: Add Phantom/Slimer CLI arguments to backstop.json config
We currently use backstop (with slimer) behind a proxy. There are CLI arguments in casper to control the proxy config but I haven't found a way to pass these parameters from backstop to casper and slimer. A sample casperjs call would be:
`casperjs --engine=slimerjs --proxy-type=http --proxy=proxyIp:port --proxy-auth=user:pass casperScript.js`

Is there a way to specify these proxy parameters (or even all casper parameters) in backstop?

paweltin:
Not sure about the code quality but it works
Closes #92


garris:
Thank you @paweltin!   Will review this week.   Best, GS

paweltin:
Tried my best with if-elses...

PR #113, opened 2015-11-12
Gives Casper scripts Viewport Info

EvanLovely:
This passes the `viewports` array from `backstop.json` into the `onReadyScript` scripts (i.e. `onReady.js`) so you can do stuff for just a specific viewport like so:

```js
module.exports = function(casper, scenario, vp) {
  casper.echo( 'Clicking button' );
  casper.click( '.toggle' );
  if (vp.name === 'phone') {
    // do stuff for just phone viewport here
  }
  // do other cool stuff here, see Casperjs.org for a full API and many ideas.
  // scenario is the current scenario object being run from your backstop.json file
}
```

Thanks for making such a great tool – I'm really loving it so far! :beers:

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,9 +317,12 @@ And in your scenario...
 Inside `filename.js`, structure it like this:
 
 ```js
-module.exports = function(casper, scenario) {
+module.exports = function(casper, scenario, vp) {
   casper.echo( 'Clicking button' );
   casper.click( '.toggle' );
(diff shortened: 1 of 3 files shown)


garris:
Nice!  That is a no-brainer!   

I am trying to get another nice feature out the door next week.  If I can manage it this will go out with it.

Cheers!

EvanLovely:
You're welcome; glad to help! :)

Issue #121, opened 2015-12-04
Save image-diffs when using the CLI reporter


klausbayrhammer:
When using the CLI reporter it is sometimes quite tricky to spot the error because the image-diffs are not persisted. As an addition to #90 it would be really useful if we store the (failing) image-diffs somewhere. We could either pre/postfix the images with passed/failed and put them in the `paths.bitmaps_test` or put them in a separat folder?

If that's ok for you I'd prepare a PR.

garris:
That makes total sense.  One nice thing about the existing CLI flow is its speed -- so for that reason it might be best to save only the failed diffs?    I like your idea of saving those in the paths.bitmaps_test/[timestamp] directory.   Maybe with a prefix (e.g. `diff_failed_`) so it is easy for a human to find when looking at the directory.   A PR is totally appreciated!   Thanks @klausbayrhammer !

garris:
merged

PR #122, opened 2015-12-04
Saves image-diffs when using the CLI reporter #121

klausbayrhammer:
Saves diff images of failed diffs in the paths.bitmaps_test/[timestamp]
 directory


garris:
Looks great!   Merged.

garris:
closes #121

Issue #117, opened 2015-11-20
Define config in JS instead of JSON


EvanLovely:
Would be great if `backstop.json` could be `backstop.js` as it'd allow the use of variables & comments. Even better would be if Backstop got config by using `require('backstop')` as that would grab any of these:

- `backstop.json`
- `backstop.js`
- `backstop/index.js`

garris:
Yes!

I just merged a configuration for setting the location of backstop.json.   I think there is plenty of room for this feature to live in harmony with that one. 

If *.js returns a promise you could easily hit an endpoint -- I think that might be the way to go.

Any case -- sounds great.  If you write it, I'll pull it.  Fine to do in small steps too.  Let me know how I can help.

Awesome.

rimi-itk:
I would like this as well and I have created a pull request: https://github.com/garris/BackstopJS/pull/125

garris:
Almost closed by PR #125 merged to master.  See documentation on how to specify config file via CLI argument.  

Would like to look at enabling backstop.js *or* backstop.json in it's default location by removing `.json` from the default config filename here https://github.com/garris/BackstopJS/blob/master/gulp/util/paths.js#L19

garris:
just looked at removing `.json` from default file -- unfortunately too many dependencies on a file extension ATM.   

Anycase,  tweaked the documentation -- pretty nice feature as is.  closing issue.  Thanks.

PR #125, opened 2015-12-17
Use "require" to load config

rimi-itk:
Feature wanted in https://github.com/garris/BackstopJS/issues/117

Must be used with --backstopConfigFilePath to use a json/js config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,16 @@ $ gulp test --backstopConfigFilePath=../../backstopTests/someTest.json
 
 NOTE: all paths are relative to the location of the BackstopJS install directory _(which is either inside your project's `node_modules` or `bower_components` depending on how BackstopJS was installed)_
 
+###Using a js based config file
+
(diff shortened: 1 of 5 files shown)


garris:
This looks good.  Can you...

1) add an example .js config file e.g. Add `test/configExample.js` 

2) briefly document for the people e.g. Add https://github.com/garris/BackstopJS#using-a-js-based-config-file

Thank you!

garris:
Come to think of it, we probably don't really need the file type check as I think require would look for either .js or .json If the suffix was left off.  I guess you can remove that check if you want.  Up to you.   Thanks again.

garris:
Oh crap, and then there is the issue of the test fails...

https://travis-ci.org/garris/BackstopJS/builds/97408007

garris:
Nailed it.  This is great -- thank you very much for the contribution!  If I get around to it I will drop the `.json` from the fallback filename to enable users to use a .js or .json file without having to specify a CLI path.   This will go out with the next release.   Cheers! :+1:

PR #128, opened 2015-12-31
Avoid publishing of test-folder

klausbayrhammer:
I wanted to remove the test directory from the published bundle and to my knowledge there are three ways to do this:

* Specifying exclusions in .gitignore
* Specify files/folders which should be published in the package.json
* Specify which files should be ignored in an .npmignore

Since we want to have the test directory in git, the first way is not an option. The second way seems to require quite high maintenance because of the nested exclusions (e.g. `capture/config.json`), so I went with the third option.

It's your call to decide whether the duplicated configs in .gitignore and .npmignore are worth that the test folder not beeing published. If you think so as well, it would be cool if you could review if we need all entries in .gitignore/.npmigore. I tried to clean them up, but there are several where I'm not sure whether we need them (e.g. `build/Release`)


garris:
Now that you have me thinking about it -- I can understand why you wanted to exclude the `test` directory.  `test` is not really a helpful name -- it is probably better named `examples` and for it to be truly helpful it should be better organized.  That would be more consistent with my comment about useful boilerplate.   If you feel it is sloppy/confusing I would agree with this and be happy to simply delete it until it can be brought back in a more organized/helpful form.

klausbayrhammer:
Good point. We could separate the mocha-specs from the example code and just publish the examples.

garris:
Yes, so in that case, it totally ok with me if you prefer leaving tests out of distribution and want to bring back the .npmignore.   Again, up to you.   Sorry for the winding road on this one.

PR #131, opened 2015-12-31
fixes #129 avoid platform specific path separator

### LINKED ISSUE #129: CLI test fails to store failed_diff images on Windows
After running `gulp test` on Windows if there were different images, process stops with error:
````
ERROR: tablets-list 6_0_body_0_xs.png
Storing diff image in  failed_diff_d:\temp\backstop-test\backstop_data\bitmaps_t
est\20151231-103406\6_0_body_0_xs.png
stream.js:74
      throw er; // Unhandled stream error in pipe.
      ^

Error: ENOENT: no such file or directory, open 'd:\temp\backstop-test\node_modul
es\backstopjs\failed_diff_d:\temp\backstop-test\backstop_data\bitmaps_test\20151
231-103406\6_0_body_0_xs.png'
    at Error (native)
````
Problem is in gulp/tasks/compare.js file, where lastSlash variable is hardcoded to support Unix file separators:
````
var lastSlash = testPath.lastIndexOf('/');
````
Shouldn't path functions be used here?

klausbayrhammer:
uses `path.sep` instead of hardcoded platform specific path separators

### DIFF
--- a/gulp/tasks/compare.js
+++ b/gulp/tasks/compare.js
@@ -65,7 +65,7 @@ gulp.task('compare', function (done) {
   }
 
   function getFailedDiffFilename(testPath) {
-    var lastSlash = testPath.lastIndexOf('/');
+    var lastSlash = testPath.lastIndexOf(path.sep);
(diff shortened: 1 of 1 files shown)


garris:
Thanks for this!  Will respond to your other issues this weekend.

Issue #124, opened 2015-12-15
Specify location of compareConfigFileName


klausbayrhammer:
Currently we are running our backstop-tests on a CI-server which operates on a fresh checkout. Since `npm install` is so painfully slow use a prepackaged node_modules directory and use `npm purge` and `npm install` to update dependencies, quite similar to [this approach](https://lincolnloop.com/blog/speeding-npm-installs/). Something we stumbled upon was that we sometimes we got the error `It looks like the reference configuration has been changed since last reference batch` although the config didn't change. In the end our problem was, that the `.config.json.cache` is stored under `node_modules/backstopjs/capture` and distributed in our packaged `node_modules.tgz` and may not match the original backstop.json. Furthermore backstop stops (with response code 0) and backstop doesn't execute the tests - which makes our CI believe that all backstop tests succeeded (but in fact they haven't been executed at all).

There are two issues we are having with the current implementation:
* If a changed config is detected the test stops without notifying the caller (like calling the gulp callback with an error or something like that)
* The cached config is placed under node_modules and the location of this cached config cannot be modified.

The easy solution for the second problem would probably be to specify the `captureConfigFileNameCache` in the `backstop.json` just like the other paths. But maybe we could come up with a more elegant solution, like computing a hash of the configuration and storing it somewhere? 

We are currently working around this by calling `bless` before we execute the tests, but with that we are loosing the config change detection (which is an awesome feature btw).
This may be quite an edge-case but I think trying to avoid storing data in the node_modules would support the implementation of #32 as well.

garris:
I understand.  I guess the two options would be 1) move the location of the cache file to the bitmaps_reference path (or an explicitly specified path in the config)
OR
2) just add a config `disableConfigCheck`  which turns off saving/checking etc..  (This is probably easiest and maybe more useful for your case)

garris:
I totally take back my last response.  Reading this on my phone on the train, I didn't see the text after the bullets.  So sorry about that!

I like your current workaround calling `bless` like that.  Totally what I would have done.

I think your first solution is the way to go.  Super straight forward.

[...]

PR #132, opened 2016-01-04
#124 uses checksum to detect config changes

klausbayrhammer:
* use npm's checksum to compute file checksums (uses sha1)
* introduces `compareConfig.lastConfigHash` to persist config checksum
* move content of `compareConfig` to `compareConfig.compareConfig`
* makes bless compute a checksum of the config and adding it to the
compareConfigFile as `lastConfigHash`

### DIFF
--- a/capture/echoFiles.js
+++ b/capture/echoFiles.js
@@ -178,8 +178,11 @@ casper.run(function(){
 });
 
 function complete(){
-  var configData = JSON.stringify(compareConfig,null,2);
-  fs.write(compareConfigFileName, configData, 'w');
+  fs.touch(compareConfigFileName);
+  var compareConfigFile = fs.read(compareConfigFileName);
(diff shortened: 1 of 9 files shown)


klausbayrhammer:
I hope I have tested all the different scenarios that can occur and so far everything seems to work as expected. But there are a few things with the current solution I'm not really satisified with:

* The changes in `echoFiles.js` and `genBitmaps.js` are nearly identical, but I'm not sure how or if you want to share code for the casper-scripts

garris:
@klausbayrhammer Great work!   

`echoFiles.js` is simply there as a sanity check when troubleshooting and should not need to write any data at all.   The fact that is still does is an oversight.  I would simply delete lines 181 & 182 from the original and leave it at that.   The file should really be cleaned up at some point.

Issue #127, opened 2015-12-29
Override defaults for misMatchThreshold/readyEvent


klausbayrhammer:
We are currently building up more and more backstop-scenarios, but since we use the same "test-setup" we are duplicating the readyEvent and misMatchThreshold in every scenario. The scenario part of our backstop config kind of  look like this:

```json
...
"scenarios": [
    {
      "label": "label1",
      "url": "url1",
      "selectors": ["selector"],
      "readyEvent": "backstop_ready",
      "misMatchThreshold" : 0.1
    },
    {
      "label": "label2",
      "url": "url2",
      "selectors": ["selector"],
      "readyEvent": "backstop_ready",
      "misMatchThreshold" : 0.1
    },
    {
      "label": "label3",
      "url": "url3",
      "selectors": ["selector"],
      "readyEvent": "backstop_ready",
      "misMatchThreshold" : 0.1
    },
]
...
```

As you can see we are duplicating the readyEvent and misMatchThreshold properties heavily and reduce the readability of our scenarios. Would it be possible to somehow specify defaults (override the defaults) for those properties. Probably the config could look like this:

```json
...
// probably this should go somewhere else or should have a different name
"defaultProperties": {  
     "readyEvent": "backstop_ready",
      "misMatchThreshold" : 0.1
}
"scenarios": [
    {
      "label": "label1",
      "url": "url1",
      "selectors": ["selector"]
    },
    {
      "label": "label2",
      "url": "url2",
      "selectors": ["selector"]
    },
    {
      "label": "label3",
      "url": "url3",
      "selectors": ["selector"]
    },
]
...
```

Probably its already possible (maybe #24) but I haven't found out how to configure it.

garris:
Yes, this should not be difficult to add.

I would not add the `defaultProperties` branch. Instead I would probably just keep the parameters (e.g. `misMatchThreshold` etc...) at the config root.   (It's ok with me if it makes more sense for your use case to group them together.)

_first, add the variables internally_

after this line...
https://github.com/garris/BackstopJS/blob/master/gulp/util/paths.js#L71
add...
```
paths.misMatchThreshold = config.misMatchThreshold || null;
paths.readyEvent = config.readyEvent || null;
```

_then, add these parameters in the app in the capture flow_

replace this line...
https://github.com/garris/BackstopJS/blob/master/capture/genBitmaps.js#L208
with...
```
misMatchThreshold: scenario.misMatchThreshold || config.misMatchThreshold
```

[...]

PR #133, opened 2016-01-05
#127 global config for misMatchThreshold/readyEvent

klausbayrhammer:
adds the ability to specify global configs for mismatchThreshold and readyEvent. Both configs can be overridden by specifying either readyEvent or mismatchThreshold on a scenario level

### DIFF
--- a/capture/genBitmaps.js
+++ b/capture/genBitmaps.js
@@ -85,9 +85,12 @@ function capturePageSelectors(url,scenarios,viewports,bitmaps_reference,bitmaps_
 
         casper.waitFor(
           function(){ //test
-            if(!scenario.readyEvent)return true;
-            var regExReadyFlag = new RegExp(scenario.readyEvent,'i');
-            return consoleBuffer.search(regExReadyFlag)>=0;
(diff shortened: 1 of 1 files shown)


garris:
I think we still need to add lines here-ish...
https://github.com/garris/BackstopJS/blob/master/gulp/util/paths.js#L71

adding...
```
paths.misMatchThreshold = config.misMatchThreshold || null;
paths.readyEvent = config.readyEvent || null;
```

garris:
Merged.  I added the paths parameters -- so I expect this to work.  Please test.  Thanks!

klausbayrhammer:
Thanks for merging. I tested it without adding the two properties to paths and it worked quite fine. `genBitmaps.js` just reads the `capture/config.json` which is a copy of the current backstop config - or am I missing something?

garris:
You are correct.   I removed the lines.   Thanks!

klausbayrhammer:
Is there a chance that you could release the last couple of changes in a version 1.0.1?

garris:
Done!

klausbayrhammer:
Awesome! Thanks!

PR #134, opened 2016-01-13
#124 makes sure tests work on fresh workspace

klausbayrhammer:
Currently the tests don't work on a fresh checkout (if no
`paths.compareConfigFileName` exists). The check whether a config has
been changed should only fail if a compareConfig exists and
contains a different checksum than the `paths.activeCaptureConfigPath`'s
checksum

### DIFF
--- a/gulp/tasks/test.js
+++ b/gulp/tasks/test.js
@@ -21,23 +21,22 @@ gulp.task('test',['init'], function () {
     genReferenceMode = true;
   }
 
-  //IF WE ARE IN TEST GENERATION MODE -- LOOK FOR CHANGES IN THE 'CAPTURE CONFIG'.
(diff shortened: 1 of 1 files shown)


klausbayrhammer:
Basically the only change is the `fs.existsSync(paths.compareConfigFileName)` before reading the config. The problem is that without this check it tries to read the `paths.compareConfigFileName` and throw an error if it doesn't exist - so the tests don't work on a fresh checkout where you have never executed a `bless` before. Sorry for the inconvenience, I was pretty sure I checked that case but it must have slipped through.
I thing we should get this out in a 1.0.2 so it doesn't cause too many troubles

garris:
Merged and published 1.0.2 to npm.    Thanks for jumping on this one so quickly!

Issue #140, opened 2016-01-19
Comparing / Diff a Production URL and a Development URL


chipcullen:
First off - great project!

I'm wondering if I'm missing something obvious: what I'm trying to do is do a comparison / diff between two different URLs - a 'production' url (http://www.example.com) and a local 'development' url (http://dev.example.com:8000).

Is there a way to make one scenario your 'reference' and one your 'test'? Or is there a different way to do this?

Thank you for any help!

garris:
@chipcullen hi, thanks for posting!  I have not considered that use case before --  it sounds like something that should be added.   Currently, BackstopJS is built around a single developer on a single URL model.   

That said, the workaround would be pretty easy.   Just set up your scenarios pointing the URLs to your reference and then run  `$ gulp reference`     Then when you want to test, just copy your config file and transform your URLs as needed (in your case, adding `:8000`) and run `$ gulp bless`  then `$ gulp test`  

( `$ gulp bless`  is to let BackstopJS know that you you've deliberately changed your config)

There are more integrated options as well.  You could make your config a node module and enable a dynamic URL based on some enviornment var.  see the docs for more info...

Hope that helps!

chipcullen:
@garris wow - thank you for the *very* fast response! 

Your work around works for now - I'll see if I can make it more dynamic.

As for the use case - I originally had seen it on Wraith - http://bbc-news.github.io/wraith/index.html#Capturemode

BUT, I'm digging an npm based project, which is why I'm trying out backstop. So far it's working great!

Is there anything I could do to help? Would a different, more "feature request" type ticket help?

garris:
Sure, I am trying to be disciplined about adding properties to the config at this point -- keep the complexity low while the project is still manageable.  That said,  I think it's a legitimate feature request.   If you wanted to do a PR I'd support it.  Should be pretty easy to add a `referenceUrl` property to a config scenario and then switch on if we are running a reference or test pass.

Replace this line...
https://github.com/garris/BackstopJS/blob/master/capture/genBitmaps.js#L84

with...
```
var url = scenario.url;
if (isReference && scenario.referenceUrl) {
  url = scenario.referenceUrl;
}
this.thenOpen(url, function() {
```
...

garris:
this is merged to master.  Would appreciate if someone could verify that the feature is working.   Thanks!

laubsauger:
[...]

PR #149, opened 2016-01-26
Updating reference echo message

chipcullen:
When running the reference, updating the echos so they don't assume scenario.url, per the discussion in #140


garris:
Nice.  Thanks!

Issue #147, opened 2016-01-26
Allow a global onReadyScript to be run throughout all scenarios


laubsauger:
First, thanks for this project :)

Second, the juicy part:

We currently have a per-scenario onReadyScript. 
A global onReadyScript would be useful for QA-use cases like generalization of dynamic content.

My specific case:
Replacing dynamic product images, texts and other data with generic placeholders. The aim is to be able to compare two versions of a site even if they are running on different datasets. These images and other elements can exist on every page. Obviously, I want to avoid duplicating code in the scenario-specific onReadyScript. Sharing a single onReadyScript throughout all scenarios is also not really desirable duer to lost flexibility.

Will provide a PR for this issue in a moment. What do you think of the idea in general?

laubsauger:
Since i was at it, i also added an onBeforeStartScript. This can be used for registering event listeners for e.g. cancelling requests or other things that have to be setup before casper starts to load the first page.

garris:
Hi @laubsauger -- thanks for posting!   

It would help me to understand why using `hideSelectors` doesn't work for you in this case.

garris:
Also, There is already a global `onReady` script option.  (It hasn't been documented yet).
https://github.com/garris/BackstopJS/pull/133

laubsauger:
Hi @garris, thanks for the hint - didn't see that. Unfortunately, the solution from #133 only allows the scenario-based onReady **or** the global onReady to run. I need both in my use case. 
The global one is used for generic scripting (e.g. replace all dynamic images, texts, etc.) while the local one caters to page specific things.

As for the onBeforeStartScript: At the moment im specifically using this to cancel requests to ads, trackers and the like. This is done to speed up things and to make tests more stable. But i'm sure there are other possible use cases.

You can see an example of this [here](https://github.com/garris/BackstopJS/blob/fafb24b0f9539d9949d4b79add44cd6028ca3400/capture/casper_scripts/onBeforeStart.global.js).

I'm definitely open to discussing theses approaches. Maybe there already is a better way to do this, but i can't see it.

garris:
Hi @laubsauger -- sorry for the delay getting back, I have been on on-call duty at work which has made it hard to respond to people.   Anycase, in your first point, I am pretty sure you can `require()` a global file in your scenario-level onReadyScript() files.  The current behavior is more of a override/fallback orientation, so assuming that `require()` will work for you, I don't think we have a strong case for making changes there.

[...]

PR #154, opened 2016-02-04
#147: added onBeforeScript; added option for global onBeforeScript/onReadyScript

laubsauger:
Implements #147


garris:
This looks beautiful!  Thank you so much -- I think this is going to be a really useful feature.

PR #161, opened 2016-02-12
Add console log colors

bradwade:
Make the CLI report a little prettier.
- Added red and green to error and passes. 
- Removed extra space. 
- Standardized on single quotes inside console.log.

### DIFF
--- a/gulp/tasks/compare.js
+++ b/gulp/tasks/compare.js
@@ -59,7 +59,7 @@ gulp.task('compare', function (done) {
 
   function storeFailedDiffImage(testPath, data) {
     var failedDiffFilename = getFailedDiffFilename(testPath);
-    console.log('Storing diff image in', failedDiffFilename);
+    console.log(' See:', failedDiffFilename);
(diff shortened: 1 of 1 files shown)


gulp/tasks/compare.js
```diff
@@ -59,7 +59,7 @@ gulp.task('compare', function (done) {
 
   function storeFailedDiffImage(testPath, data) {
     var failedDiffFilename = getFailedDiffFilename(testPath);
-    console.log('Storing diff image in ', failedDiffFilename);
+    console.log('Storing diff image in', failedDiffFilename);
```
garris: As long as we're at it, maybe the text should be shortened to: 
"       See: "

Essentially indent the file info to start with the error data value.

bradwade: Good idea. Will update PR.

bradwade:
I indented the label three spaces which is sufficient to show hierarchy, economizes space (I have some looong file names), lines up the colons, and places the file paths at the beginning of the error message.

garris:
Looks good -- merged -- thanks!

PR #162, opened 2016-02-13
Separate creating browser report data from starting server and launching browser.

bradwade:
This PR...
- separates into two tasks creating browser report data and the launching of the server/browser.
- creates a report option to only create report data.
- maintains backwards compatibility, so all commands continue to work as expected.
- updates documentation.
- only starts the server when the browser will be opened. (It appears that previously the server would start even when only the CLI report was being run.)

Use case: running headless on a local/dev server to prepare reports that can be seen later through the dev server's already running apache/nginx server. I'm thinking about setting up a Jenkins job to automatically run a backstop test overnight that can be viewed by QA team and devs each morning.

### DIFF
--- a/gulp/tasks/report.js
+++ b/gulp/tasks/report.js
@@ -1,10 +1,14 @@
 var gulp = require('gulp');
 var paths = require('../util/paths');
 
-gulp.task('report',['start'],function(){
+gulp.task('report', function(){
 
   if (!paths.report || paths.report.indexOf( 'browser' ) > -1 ){
     setTimeout(function(){gulp.run('openReport')},100);
+  }
+
+  if (!paths.report || paths.report.indexOf( 'prepare' ) > -1 ){
+    setTimeout(function(){gulp.run('createReport')},100);
   }
 
   if (!paths.report || paths.report.indexOf( 'CLI' ) > -1 ){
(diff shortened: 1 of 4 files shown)


garris:
Nice work!  I'm going on holiday -- i can give this a proper review in a few days.  will follow up soon.  Cheers.

klausbayrhammer:
I'm really looking forward to this feature! Thanks a lot @bradwade. I'd love to use the prepare-report it in combination with the CLI-Report., so in the end I think it's quite a cool feature to support multiple reports for the same test run.
I think it would be a good idea to just specify a single report as default (probably `browser`). We could also try to catch conflicting configurations (like `prepare` and `browser`) and stop the test or just execute `browser` if both are specified since it is a superset of `prepare`. I like the second option better.
In addition: What do you think about another name for the new report: something like `html`

gulp/tasks/report.js
```diff
@@ -1,12 +1,16 @@
...
   if (!paths.report || paths.report.indexOf( 'browser' ) > -1 ){
     setTimeout(function(){gulp.run('openReport')},100);
   }
 
+  if (!paths.report || paths.report.indexOf( 'prepare' ) > -1 ){
+    setTimeout(function(){gulp.run('createReport')},100);
+  }
+
   if (!paths.report || paths.report.indexOf( 'CLI' ) > -1 ){
     setTimeout(function(){gulp.run('compare')},1000);
```
klausbayrhammer: I'm not sure why we need the timeouts for the CLI/prepare reporters. I'm pretty sure we can run them without delay.

garris:
Closed by #259

PR #163, opened 2016-02-19
Use flexbox to size column widths on the BackstopJS report

jcdarwin:
Use flexbox to size column widths on the BackstopJS report.

Previously, using a table with 4 columns at 25% width didn't always results in equal columns.

This meant that the diff image would not necessarily line up with the reference and test images meaning it was difficult on long pages to easily eyeball the identified differences with the reference and test images.

Using flexbox means that we can guarantee that the columns will be of equal width, and hence the images will line up properly.

### DIFF
--- a/compare/css/styles.css
+++ b/compare/css/styles.css
@@ -7,8 +7,25 @@ td img{width: 100%;}
 table{width: 100%;border-collapse: collapse;margin-bottom: 50px;}
 input{width: 55ex; display: none;}
 th{text-align: left;background: #ddd;padding: 10px;white-space: nowrap;}
-th.selector,th.filename{background-color: #666;color:#fff;}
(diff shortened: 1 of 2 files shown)


garris:
Could you let me know what browsers have you tested this with?

jcdarwin:
Yep, but found one small problem with Firefox -- have rectified in the
latest PR:

https://github.com/garris/BackstopJS/pull/166

Note that Flexbox support is pretty good these days, but IE9 and before are
not supported:
http://caniuse.com/#feat=flexbox

Cheers,

Jason

garris:
Great.  Merged.   So,  Chrome + Safari + FF should be good now?

jcdarwin:
Yep, and also Edge and IE10+

Jason

PR #169, opened 2016-03-08
Update PhantomJS

wjramos:
- Add phantomjs2 and casperjs to package.json as dependencies
- Update gulp
- Replace references to deprecated phantomjs package


garris:
@wjramos -- thank you for this PR!   I have been wanting to upgrade the PhantomJS references.

Couple of things though...

- this PR breaks the build -- please review https://travis-ci.org/garris/BackstopJS/builds/114450618

- I don't believe that adding phantom as a local dependency will necessarily work.   For example, imagine you do a global install of phantom 1.9.8 and then install BackstopJS with your modified package.json.  You'll have 2 versions of phantom on your machine.  Which one will casperjs use?  I am pretty sure it will use the global one.  You can test this by adding a custom backstop script that logs out the phantom version when run.  To ensure the local version is run, I believe you'll need to add a line like this...

     ```
     process.env["PHANTOMJS_EXECUTABLE"] ="./node_modules/phantomjs/bin/phantomjs"
     ```
     before this line...
     https://github.com/garris/BackstopJS/blob/master/gulp/tasks/test.js#L66

- the same is also probably true for casperjs.   I am pretty sure you'll need to `import('./node_modules/casperjs/bin/casperjs')` anywhere casperjs is imported.

It would be AWESOME if you could get all of this to come together.   I simply don't have the time to implement and test this.   Most of what you'll need to make everything work is in the post -- but feel free to reach out with questions.

Best,
Garris

wjramos:
Thanks @garris for the prompt response.

I will need to look into how we are referencing phantom and casper, but it should always pick up the local or whatever is in the closest parent directory. If the same version is installed globally npm install should just link it.

It looks like I added an extra comma by mistake -- it's rebuilding now, we'll see how it goes without a syntax error.

garris:
@wjramos I was working on migrating from `gulp <command>` to `npm run <command>` today -- which allows us to eliminate the global Gulp reference -- while I was at it I tested out your scenarios above and confirmed you were completely correct!   So, I want to thank you for your PR -- it gave me good perspective and enabled me to completely eliminate all globally installed dependencies.

I didn't check if the versions I used were as up to date as yours so please feel free to take a look again at the package.json and update where necessary.

wjramos:
@garris So, just to clarify, gulp is no longer going to be a dependency at all?

garris:
Hi @wjramos.  Gulp is still used locally.  The last update eliminates the need for the additional _global_ install of gulp.  

Eventually, we'd probably remove gulp altogether.

garris:
Also thanks for the update on phantomjs2 -- I had been waiting so long for that to come out that I pretty much decided it would never happen. ;)

axe312ger:
Haha yes, garris, i had the same tough back then when we added slimerjs.

You guys should consider using this NPM wrapper for phantomjs. This will a lot of time while preparing the tests, especially now when using test runners is starting to get really useful here :)

--> https://www.npmjs.com/package/phantomjs-prebuilt

garris:
@axe312ger   Yup -- that's the PhantomJS version we are on.  Install is much easier now!

axe312ger:
okay sweet, was just checking the change in this PR, thats why i came up with prebuild :)

jebenois:
What is the status with using Phantomjs2? 

The docs seems to indicate Phantomjs2 is not being used: 
"PhantomJS, the default rendering engine, does not correctly interpret flexbox and web fonts -- so if you are using those things in your app you will be way more happy using SlimerJS. Here is how to do that..."

garris:
BackstopJS is using Phantom 2  -- we didn't use this pull request verbatim but did update the reference.   We still need to update the documentation -- so thanks for letting us know!

PR #170, opened 2016-03-09
prevent clipping when using "body" selector

medovob:
When using a "body" selector, the capture is clipped to the area of the viewport.

Using `casper.capture()` you can capture the entire document regardless of size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,6 +590,11 @@ From `./node_modules/backstopjs` ...
 $ gulp echo
 ```
 
+#### Body capture is clipped  
+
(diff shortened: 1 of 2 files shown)


medovob:
Actually, I think this might only occur if the document body is set to `height:100%`

In any case, I think it would be good to have this as an option at least

garris:
This would almost certainly break some existing tests so I'd rather not spring this on users in quite this way. Why not switch on a reserved value like `body:noclip`?   Also, please update the documentation to let others know about the feature.

garris:
Looking good.  will test soon.  Thanks!

fillerwriter:
FWIW, I was able to use the code from this PR against lastest master without issue. Thanks @medovob!

garris:
Many apologies for the delay getting this in.   And many many thanks for the contribution! @medovob @fillerwriter

PR #181, opened 2016-04-19
include wait in example custom casper script

malcomio:
it isn't immediately obvious that it's necessary to include a wait inside custom scripts for any page state changes to be included in the screenshots, so I'd suggest including this in the example


garris:
This is a great point -- many thanks for the PR!

PR #137, opened 2016-01-14
Add possibility to set filename.

Primajin:
#### Setting the names for generated files

By default the file names follow a certain pattern. You can override this by setting the `fileName` property in your `backstop.json` to something other than `null`.
Use these tokens and remember to create unique files and set the file extension to `.png`

| Token | Replacement |
| ----- | ----------- |
| `__scenario_index__` | **0**_0_header_0_phone.png |
| `__counter__` | 0_**0**_header_0_phone.png |
| `__cleanedSelectorName__` | 0_0_**header**_0_phone.png |
| `__viewport_index__` | 0_0_header_**0**_phone.png |
| `__vp.name__` | 0_0_header_0_**phone**.png |

default sequence is:

```js
scenario_index + '_' + i + '_' + cleanedSelectorName + '_' + viewport_index + '_' + vp.name + '.png'
```

PR #197, opened 2016-06-05
Add possibility to set filename.

Pofigizm:
I see that you haven't progress to finish #137 
I tried to do it.

I revert all changes which changed default flow and add simple docs.


Pofigizm:
@garris, is PR helpful?

Pofigizm:
It was added in https://github.com/garris/BackstopJS/pull/247

PR #208, opened 2016-06-24
Update README.md

imclapton:
Add info about misMatchThreshold property.


levithomason:
I think this is already in the README.md:

>### Changing test sensitivity
>
>`"misMatchThreshold"` will change the amount of difference BackstopJS will tollerate before marking a test screenshot as "failed". The default setting is `0.1`, this may need to be increased as the dimentions of your screenshot increases.

https://github.com/garris/BackstopJS#changing-test-sensitivity

garris:
Thanks @imclapton and @levithomason.   The readme is updated.

PR #209, opened 2016-06-25
custom testSuiteName fix

nobso:
The custom testSuiteName does not work and it's always taking the default name which is 'BackstopJS'. This would fix and make it work as expected.


nobso:
@garris could you please review this?

garris:
LGTM.

garris:
I just put an untested feature on master.   I've asked the requesters to verify.  If they can do it in the next few hours we can release the whole thing tonight.

nobso:
Thanks! @garris

Issue #210, opened 2016-06-27
Having trouble generating ci_report


mousemke:
I'm trying to generate ci reports to integrate with jenkins, but whatever i try, i cant get it to output a file.  I've added "CI" to the `json`, as well as changing the directory repeatedly.  it wont generate a report and doesnt throw an error

vague prob, i know, but i can provide code if necessary.

garris:
What version are you using?

mousemke:
`^1.3.2`

garris:
@mousemke could you please post your config file and any other info here. 

@nobso Since this is your feature, would you mind taking a quick look?

mousemke:
mostly default https://gist.github.com/mousemke/77e08417d3c6b3852481b75d51e89a3c

but i've toyed with plenty of options (setting the directories, doing `["CI"]` only, etc)

nobso:
@mousemke, Where have you been looking for a generated file? The report will be generated when you run 'npm run test'.

Can you please add the path for CI report and look at the path?

```
"paths": {
       "ci_report" :  "../../backstop_data/ci_report"
},
```

mousemke:
still the same result.  no ci_report anything (searched whole drive, for the sake of it)

also for the sake of it, my package.json script
`"test:visual:cli": "npm run test:visual:clean && cd node_modules/backstopjs/ && npm run test"`

mousemke:
output (report changed to just "CI" for brevity.  no mention of ci, no errors)

```
➜  backoffice git:(master) ✗ npm run test:visual:cli

> backoffice@0.0.1 test:visual:cli /Library/WebServer/Documents/backoffice
> npm run test:visual:clean && cd node_modules/backstopjs/ && npm run test


> backoffice@0.0.1 test:visual:clean /Library/WebServer/Documents/backoffice
> cd node_modules/backstopjs/ && rm -rf bitmaps_test


> backstopjs@1.3.2 test /Library/WebServer/Documents/backoffice/node_modules/backstopjs
> gulp test


BackstopJS Config loaded at location /Library/WebServer/Documents/backoffice/backstop.json
(node:4503) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
[18:23:44] Using gulpfile /Library/WebServer/Documents/backoffice/node_modules/backstopjs/gulpfile.js
[18:23:44] Starting 'init'...
[18:23:44] Finished 'init' after 90 μs
[...]

PR #211, opened 2016-06-27
fixing the CI report for mutiple CSS selectors

nobso:
The ```compareConfig.testPairs``` would always be an Object and not an Array. This would fix the https://github.com/garris/BackstopJS/issues/210


nobso:
I did one round of testing and it works fine for me. It would be great if @mousemke confirms that this change fixes the issue. So that @garris would be able to merge it.

garris:
Thank you very much for addressing this so quickly -- much appreciated! 👍

nobso:
@garris Thanks! 😄

PR #212, opened 2016-06-27
added a files array to reduce npm bloat

mousemke:
just a small tweak so `npm i` doesn't install dev and www files


axe312ger:
@garris thats a quick win :)

PR #216, opened 2016-07-01
test report file generation [Fixes #214]

### LINKED ISSUE #214: Xunit report file is not generating correctly
When running tests the xunit report file is not generating correctly.
I have tracked down the issue to the comapre.js file.
It generated the report file after comparison and only if the key equals testPairsLength + 1 however the comparisons are not run in the same order so the last pair will not always be the last one to run which leads to incorrect data written in the file.
If I move that block for writing the file in function updateProgress just bellow the "Test completed" messages then the data of the xunit file is correct and has all tests data.
If you want I can make PR with that code change.

nobso:
Please find the bug details #214

### DIFF
--- a/gulp/tasks/compare.js
+++ b/gulp/tasks/compare.js
@@ -24,6 +24,13 @@ gulp.task('compare', function (done) {
             console.log ('\x1b[32m', (results.pass || 0) + ' Passed', '\x1b[0m');
             console.log ('\x1b[31m', (results.fail || 0) + ' Failed\n', '\x1b[0m');
 
+            // if the test report is enabled in the config
+            if (testSuite) {
+                junitWriter.save(path.join(paths.ci_report, 'xunit.xml'), function() {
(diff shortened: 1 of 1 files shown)


nobso:
@garris I did one round of testing in my local system, it seems to be working fine. Can you please review this?

@vflirt please confirm if that works for you too.

levithomason:
If you update this PR description just a tad, you can [close the related issue when this pr is merged](https://github.com/blog/1506-closing-issues-via-pull-requests).

nobso:
@levithomason commit message would also do the same

levithomason:
Indeed, missed that!

garris:
This looks good to me -- thanks @nobso!

I'm going to go ahead and merge but would love to hear from @vflirt to validate if possible.

Cheers!

nobso:
@garris @vflirt  Thank you for making it better!

PR #220, opened 2016-07-08
Fix referenceUrl bug when using onBeforeScript

RickyFullBottle:
Previously, if you use referenceUrl which differs from url. the scenario.url object is still using url. 
in the case below with an onBeforeScript to login: `capser.thenOpen(scenario.url)` will ignore the referenceURL and still use the test url. 

    // EXAMPLE: LOGIN BEFORE RUNNING TESTS
    module.exports = function(casper, scenario, vp) {
      casper.thenOpen(scenario.url, function(){
         if (this.exists('form#user-login-form')) {
           this.fill('form#loginForm',{
              'username': 'test',
              'password': 'changeme'
           }, true);
         }
      });
    };

### DIFF
--- a/capture/genBitmaps.js
+++ b/capture/genBitmaps.js
@@ -75,10 +75,10 @@ function capturePageSelectors(url,scenarios,viewports,bitmaps_reference,bitmaps_
         this.viewport(vp.width||vp.viewport.width, vp.height||vp.viewport.height);
       });
 
-      var url = scenario.url;
       if (isReference && scenario.referenceUrl) {
-        url = scenario.referenceUrl;
+        scenario.url = scenario.referenceUrl;
       }
+      var url = scenario.url;
 
       var onBeforeScript = scenario.onBeforeScript || config.onBeforeScript;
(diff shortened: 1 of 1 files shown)


garris:
@rickyfullbottle thank you very much for posting this issue.  I think you are on the right track with this however,  I am not sure that it's a good idea to clobber a config parameter -- as there may be a process downstream that is dependent on the intrinsic `scenario.url` value.    Instead of modifying the url parameter, would you consider adding a `scenario.isReference` so that a user script could be programmed to behave as desired?   Alternatively, you could add a `scenario.currentUrl` which would return the correct url.

What are your thoughts?

RickyFullBottle:
@garris If there is a referenceUrl, then shouldn't all the url be set to referenceUrl downstream anyway? Isn't it the same as changing the url to another link that used for running reference?

RickyFullBottle:
actually maybe a better question would be, is there a case which url should not equal to referenceUrl in referenceUrl is present while generating the reference images?

PR #230, opened 2016-07-25
Use of `npm explore` to run `backstopjs` commands

gtramontina:
From the docs (https://docs.npmjs.com/cli/explore):

> Spawn a subshell in the directory of the installed package specified.
> If a command is specified, then it is run in the subshell, which then immediately terminates.


garris:
Closed by #259

PR #244, opened 2016-08-09
Pass isReference value to onReady and onBeforeReady scripts

natalan:
This is a follow up PR to https://github.com/garris/BackstopJS/pull/220

Instead of mutating `scenario` object, let's add fourth argument to both `onReady` and `onBeforeReady` scripts.


natalan:
Closing this PR to reopen on a different branch

PR #246, opened 2016-08-09
Pass isReference value to onReady and onBeforeReady scripts

natalan:
This is a follow up PR to #220

Instead of mutating scenario object, let's add fourth argument to both onReady and onBeforeReady scripts.


garris:
That makes total sense -- very easy to grok!      

Merged.   Thank you!

PR #247, opened 2016-08-10
Add possibility to set filename.

Primajin:
Refers to #137 

@garris If you want the commits squashed I can do that beforehand, but it would also be possible to just use the github squash Button:

Image of Squash and Merge

https://help.github.com/articles/about-pull-request-merge-squashing/

Which one would you prefer?

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,6 +439,27 @@ Please note: these file paths are relative to your `./node_modules/backstopjs/` 
   }
 ```
 
+### Setting custom reference file names (version 1.3.0+)
+By default, BackstopJS saves its screenshots using the next file name template
+
+```
+{scenarioIndex}_{selectorIndex}_{selectorLabel}_{viewportIndex}_{viewportLabel}.png
+```
+
+You can add your own template in config file
+
+```json
+  "fileNameTemplate": "{configFile}_{scenarioIndex}_{scenarioLabel}_{selectorIndex}_{selectorLabel}_{viewportIndex}_{viewportLabel}"
+```
+
+You can also set human name for your config which will replace name (or hash) of your `configFile`
+
+```json
+  "name": "Default"
+```
+
+NB! All symbols except `0-9`, `a-Z`, `-` and `_` will be removed in reference file names.
+
 ### Changing the rendering engine (version 0.6.0+)
 BackstopJS supports using PhantomJS or SlimerJS for web app rendering. (With thanks to CasperJS for doing the heavy lifting here.)
 
--- a/capture/genBitmaps.js
+++ b/capture/genBitmaps.js
@@ -18,6 +18,8 @@ var casper_scripts = config.paths.casper_scripts || null;
 var compareConfigFileName = config.paths.compare_data || 'compare/config.json';
 var viewports = config.viewports;
 var scenarios = config.scenarios||config.grabConfigs;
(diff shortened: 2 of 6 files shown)


garris:
I forgot about that squish feature -- thanks!  Will try to run this today.

Primajin:
🎉  Great!

README.md
```diff
@@ -439,6 +439,27 @@ Please note: these file paths are relative to your `./node_modules/backstopjs/`
   }
 ```
 
+### Setting custom reference file names (version 1.3.0+)
```
Pofigizm: Don't forget change to correct version here.

Primajin:
Thanks @pofigizm for the reminder....actually I lost  a bit track of the version counting...maybe @garris can help out here and say when its time to go for major, minor or patch.

😉

garris:
@Primajin thanks so much for the feature work!  I had just one comment which I left on the PR.   please just address that and this feature will be merged for the next release.  

FYI: there is another feature #249 (selector expansion) which would effect behavior of the filenaming (downstream of your naming template).   Please feel free to make any comments, suggestions etc as this would dynamically add selector names for selectors with multiple instances.

Cheers.

Primajin:
Hey @garris , great! Just one thing

> I had just one comment which I left on the PR. please just address that [...]

Sorry I don't get what you mean. Where is this comment? Can you point me there. Is it in #137 or in here? Is it a line comment or a PR note? I will work on that asap and will also have a look at #249 !

Best
Jannis / @Primajin

capture/genBitmaps.js
```diff
@@ -12,12 +11,25 @@ if (!config.paths) {
   config.paths = {};
 }
 
+String.prototype.hashCode = function () {
```
garris: What is the reason for adding this to the string prototype as opposed to just a local method?

garris:
@Primajin sorry, the comment is here ^^^

capture/genBitmaps.js
Primajin: Well it was "lented" from here: http://werxltd.com/wp/2010/05/13/javascript-implementation-of-javas-string-hashcode-method/ ;)

garris:
Just an update.  I am going to merge this into the version_2_0_RC branch.  Version 2 is a huge refactor (completely removed gulp) and this feature is required for that one.   I'll ping everyone when it's in to give people a chance to test.   Thank you very much for this feature work  -- I appreciate your persistence.   Cheers.

onigoetz:
By the way, we already changed the default file name in this new branch : https://github.com/onigoetz/BackstopJS/blob/1f8d7a6c95264dbf5f95fbf5d37c6bacf5405413/capture/genBitmaps.js#L173

We use the label of a scenario to generate the name and not the scenario's index.

matthew-dean:
How do you actually change the reference file name using this feature? o_O I can't find any existing documentation to do so.

garris:
It is quite easy — however — not recommended since backstop requires enough unique information in the filename to manage everything.    That said, if you’re careful you can probably get it to work just fine.

Any case, check the readme here... https://github.com/garris/BackstopJS/blob/master/README.md#if-you-just-upgraded-to-2x-or-3x

matthew-dean:
Yep, I figured it out but forgot to find this thread to update you, sorry about that.

PR #248, opened 2016-08-10
support config opening report domain url

JackZhang1988:
When i installed BackstopJs in my linux server, i want to usr my linux server domain url as the report url so that when i click the console output, i can directly redirect to the compare server.


garris:
Makes sense.  Merged.  Thanks!

PR #249, opened 2016-08-11
one-to-many selector testing capability is added [Fixes #245]

### LINKED ISSUE #245: single reference screenshot for multiple/same elements
Hi @garris ,
I've a reference screenshot for an HTML component (0_0_buttonclass_0_desktop.png) that i need to match against multiple components (buttons) that have the same CSS class names. For now, only the very first component is getting matched and others are ignored. 

I don't want to duplicate the reference screenshot as the expectation of other button components are going to be same. But i want to test it to make sure they are same.

Do we any provision for this one?

nobso:
In the current version, there is only one-to-one selector testing can be done. All other similar selector elements are ignored after the first match.

This PR will add the capability to run tests on many similar selectors against one reference screenshot. 

This is not a configuration driven feature but we can make it if needed.

@garris what do you think of this? Please leave your comments if you have any.

### DIFF
--- a/capture/genBitmaps.js
+++ b/capture/genBitmaps.js
@@ -148,6 +148,7 @@ function capturePageSelectors(url,scenarios,viewports,bitmaps_reference,bitmaps_
               scenario.removeSelectors.forEach(function(o,i,a){
                 casper.evaluate(function(o){
                   Array.prototype.forEach.call(document.querySelectorAll(o), function(s, j){
+                    s.classList.add('_nIgnore');
                     s.style.display='none';
                   });
                 },o);
@@ -159,16 +160,41 @@ function capturePageSelectors(url,scenarios,viewports,bitmaps_reference,bitmaps_
             if ( !scenario.hasOwnProperty('selectors') ) {
               scenario.selectors = [ 'body' ];
             }
+
+        // if similar selectors exist on the given page, expand and test them too
+        if (scenario.enableSelectorExpansion) {
+            scenario.selectors = scenario.selectors.reduce(function(acc, selector, index) { 
+                var expandedSelectors = casper.evaluate(function(selector, index) {
+                    return [].slice.call(document.querySelectorAll(selector)).filter(function (element) {
+                        
+                        // dont test if they are added into the removeSelectors config section
+                        return !~(element.classList + '').indexOf('_nIgnore');
+                    }).map(function(element, i) {
+
+                            // by using __backstopjs__, we are making sure that this classname does not clash with user defined ones 
+                            var nPart = '__backstopjs__' + index + '__' + i;
+                            element.classList.add(nPart);
+                            return selector + '.' + nPart;
+                    });
+                }, selector, index);
+                return acc.concat(expandedSelectors);
+            }, []);
+        }
+
+
         scenario.selectors.forEach(function(o,i,a){
           var cleanedSelectorName = o.replace(/[^a-z0-9_\-]/gi,'');//remove anything that's not a letter or a number
           //var cleanedUrl = scenario.url.replace(/[^a-zA-Z\d]/,'');//remove anything that's not a letter or a number
+          
+          // make sure that we always refer the actual reference file & not the expanded ones
+          if (/__backstopjs__\d*__\d*/.test(cleanedSelectorName)) {
+            i = cleanedSelectorName.match(/__backstopjs__(\d*)__\d*/)[1];
+          }
           var fileName = scenario_index + '_' + i + '_' + cleanedSelectorName + '_' + viewport_index + '_' + vp.name + '.png';;
-
-          var reference_FP  = bitmaps_reference + '/' + fileName;
+          var reference_FP  = bitmaps_reference + '/' + fileName.replace(/__backstopjs__\d*__\d*/, '');
           var test_FP       = bitmaps_test + '/' + screenshotDateTime + '/' + fileName;
 
           var filePath      = (isReference)?reference_FP:test_FP;
-
 
           if(o === "body:noclip" || o === "document") {
             casper.capture(filePath);
@@ -184,8 +210,7 @@ function capturePageSelectors(url,scenarios,viewports,bitmaps_reference,bitmaps_
             fs.write(filePath, assetData, 'b');
           }
 
-
-          if (!isReference) {
+        if (!isReference) {
             compareConfig.testPairs.push({
               reference:reference_FP,
               test:test_FP,
--- a/test/selectorExpansion/backstop.json
+++ b/test/selectorExpansion/backstop.json
@@ -1,0 +1,43 @@
+{
+  "viewports": [
+    {
+      "name": "phone",
+      "width": 320,
+      "height": 480
+    },
+    {
+      "name": "desktop",
+      "width": 1440,
+      "height": 900
+    }
+  ],
+  "scenarios": [
+    {
+      "label": "primary buttons",
+      "url": "index.html",
+      "hideSelectors": [],
+      "removeSelectors": [
+        ".container .btn.btn-primary.anyclassname"
+      ],
+      "selectors": [
+        ".container  .btn.btn-primary"
+      ],
+      "enableSelectorExpansion": true,
+      "delay": 100,
+      "misMatchThreshold" : 0.1,
+      "onBeforeScript": "onBefore.js",
+      "onReadyScript": "onReady.js"
+    }
+  ],
+  "paths": {
+    "bitmaps_reference": "../../backstop_data/bitmaps_reference",
+    "bitmaps_test": "../../backstop_data/bitmaps_test",
+    "compare_data": "../../backstop_data/bitmaps_test/compare.json",
+    "casper_scripts": "../../backstop_data/casper_scripts"
+  },
+  "engine": "phantomjs",
+  "report": ["CLI", "browser"],
+  "casperFlags": [],
+  "debug": false,
+  "port": 3001
+}
--- a/test/selectorExpansion/index.html
+++ b/test/selectorExpansion/index.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+        <meta name="description" content="">
+        <meta name="author" content="">
+        <title>selectorExpansion - backstopJS</title>
+        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
+    </head>
+    <body>
+        <div class="container">
+            <div class="starter-template">
+                <h1>Example: selector expansion</h1>
+                <p class="lead">There are 3 buttons with the same CSS classname (.btn-primary). If you want your test script to expand and test all the similar CSS components (.btn-primary), just use <code>enableSelectorExpansion: true</code> in the backstop config as mentioned in the example config file.</p>
+                <input class="btn btn-primary" type="submit" value="Submit">
+                <input class="btn btn-primary anyclassname" type="submit" value="Submit">
+                <input class="btn btn-primary" type="submit" value="Submit">
+            </div>
+        </div>
+    </body>
+</html>


nobso:
@garris May i know if you had a chance to look at this requirement?

capture/genBitmaps.js
```diff
@@ -169,12 +170,48 @@ function capturePageSelectors(url,scenarios,viewports,bitmaps_reference,bitmaps_
...
+
+                        // let's add the temporary CSS class name to group the similar elements
+                        element.classList.add('backstopJS' + index); 
+                    });
+                    return allSimilarElements.length;
+                }, o);
+
+                // if there are more elements that has similar CSS seletors
+                if (targetElementsLength > 1) {
+                    var insertPos = fileName.split('_', 2).join('_').length,
```
garris: There is another feature branch where someone is working on a custom filename #197.  If we move forward with this we would need to keep that other feature in mind -- probably by extending that feature.

nobso: Sure. I will take a look at #197 and come back, so that the discussion would make sense.

capture/genBitmaps.js
```diff
@@ -169,12 +170,48 @@ function capturePageSelectors(url,scenarios,viewports,bitmaps_reference,bitmaps_
...
+                        testFilesDirectory = filePath.substr(0, filePath.lastIndexOf('/') + 1),
+                        testFileName,
+                        testFilePath,
+                        elementIndex;
+
+                    // take screenshots of all the elements that use the same selector
+                    for (elementIndex = 0; elementIndex < targetElementsLength; elementIndex++) {
+                        testFileName = fileName.substr(0, insertPos) + '_backstopJS' + elementIndex + fileName.substr(insertPos);
```
garris: Maybe use a shorter delim here?   Maybe just an `_n`?

nobso: These new filenames are auto generated and users should not get confused with their names, so wanted to let them know explicitly that they are created by backstopJS.

But i can change to '_n' and update the README accordingly.

capture/genBitmaps.js
nobso: @garris I've covered your concern about #197 in this PR. Please review

nobso:
@garris i think, you might be busy with other stuffs, but it would be great if if you take re-look at it

garris:
Apologies, I have been traveling.  Will try to review this today -- but may not have a chance till I return on Thursday.   Thank you for your patience.

nobso:
@garris No problem. Thanks for the update

garris:
Ok, I have reviewed the changes and have some thoughts and suggestions.

First, I really like this feature.  Thank you for your inspired contribution here.

I think an appropriate name for this would be something like "selector expansion".  I think this feature should also be switchable, e.g. `enableSelectorExpansion: true` as a scenario-level config.  And this should default to true.    

The current implementation has some issues.  It appears the (new) "expanded" instances would not handle missing or invisible element test cases in the same way that the (current implementation's) "first-of-type" or 1:1 selectors do.  E.g. See the following two blocks for examples of missing/hidden case handlers outside the scope of the new expanded selectors...
https://github.com/nobso/BackstopJS/blob/c1b00863f8e2509544a380ed49c90e6d6ca7892f/capture/genBitmaps.js#L231

https://github.com/nobso/BackstopJS/blob/c1b00863f8e2509544a380ed49c90e6d6ca7892f/capture/genBitmaps.js#L235

I think this implementation may be cleaner if inserted before the existing flow such that no changes to the existing selector loop would be needed.  Have you considered something like flat-mapping the selectors into a new array?

Replace this line...
https://github.com/garris/BackstopJS/pull/249/files#diff-8854ff0ca17d960be935d14ef5decc3fR163

With something like this...
```
if (scenario.enableSelectorExpansion === undefined || scenario.enableSelectorExpansion) {
scenario.expandedSelectors = scenario.selectors.reduce(function(acc, selector){ 
  var expandedSelectorArray = // awesome selector targeting and naming code here...
  return acc.concat(expandedSelectorArray);
}, []);
else {
  scenario.expandedSelectors = scenario.selectors
}

scenario.expandedSelectors.forEach(function(o,i,a){
```
With this approach, I think you would avoid the need for the `ignore` meta-classes.

What do you think?

garris:
@nobso ^^^

garris:
@nobso   I have reviewed the changes and have some thoughts and suggestions.

First, I really like this feature.  Thank you for your inspired contribution here.

I think an appropriate name for this would be something like "selector expansion".  I think this feature should also be switchable, e.g. `enableSelectorExpansion: true` as a scenario-level config.  And this should default to true.    

The current implementation has some issues.  It appears the the (new) "expanded" instances would not handle missing or invisible element test cases in the same way that the (current implementation's) "first-of-type" or 1:1 selectors do.  E.g. See the following two lines for examples of missing/hidden case handlers outside the scope of the new expanded selectors...
https://github.com/nobso/BackstopJS/blob/c1b00863f8e2509544a380ed49c90e6d6ca7892f/capture/genBitmaps.js#L231

https://github.com/nobso/BackstopJS/blob/c1b00863f8e2509544a380ed49c90e6d6ca7892f/capture/genBitmaps.js#L235

I think this implementation may be cleaner if inserted before the existing flow such that no changes to the existing selector loop would be needed.  Have you considered something like flat-mapping the selectors into a new array?

Replace this line...
https://github.com/garris/BackstopJS/pull/249/files#diff-8854ff0ca17d960be935d14ef5decc3fR163

With something like this...
```
if (scenario.enableSelectorExpansion === undefined || scenario.enableSelectorExpansion) {
scenario.expandedSelectors = scenario.selectors.reduce(function(acc, selector){ 
  var expandedSelectorArray = // awesome selector finding and naming code here...
  return acc.concat(expandedSelectorArray);
}, []);
else {
  scenario.expandedSelectors = scenario.selectors
}

scenario.expandedSelectors.forEach(function(o,i,a){
```
With this approach, I think you could avoid the need to decorate the DOM with meta-classes.

What do you think?

garris:
Also, to be backward compatible from a behavior standpoint, the first of all expanded selectors found should have no name extension e.g. `.myLogo`.  But subsequent similar elements should be decorated with an additional selector class `myLogo_n1`, `myLogo_n2`.

garris:
@nobso I think this line from your post above may be confusing to some other contributors...

> This PR will add the capability to run tests on many similar selectors against one reference screenshot.


I think what you mean is this...

_This PR will add the capability to run tests on many instances of a single selector._

Just wanted to confirm this and be certain that is your intent...  Is that true?

nobso:
@garris thanks for your comments. I will look into it this weekend. 

> This PR will add the capability to run tests on many instances of a single selector.

what do you mean by 'many instances'? I meant 'many occurrences' of a same element on the given page.

For example,

If we have a reference screenshot for ``` .author-profile-image``` , we should be able to compare all the elements (You may not know how many of them beforehand) that have 'author-profile-image' class against the reference screenshot. Right now, its compares only the first one.

``` HTML
<div class="author-profile-image">IMAGE</div>
<div class="author-profile-image">IMAGE</div>
<div class="author-profile-image">IMAGE</div>
```

garris:
Ok good.  Thank you for the clarification-- I think there was just some confusion with the language of that line.   

Please carry on.  🙂

garris:
@nobso I just wanted to alert you to a very positive development happening on the `version_2_0_onigotez` branch.   The BackstopJS core has been rewritten and Gulp has been completely removed.   Your change (which will be a great feature!) would still move forward and remain the same -- it would just be happening in a different file and slightly different context.   

I wanted to make you aware of the forthcoming work in progress please feel free to make comments and suggestions etc.

Hope you are having a good weekend.

Cheers.

nobso:
@garris sorry for delaying this. This week, i will try to work on it.

garris:
No worries at all!  Thank you.

nobso:
> The current implementation has some issues. It appears the the (new) "expanded" instances would not handle missing or invisible element test cases in the same way that the (current implementation's) "fir ....

I think, we may not need to worry about the hidden/invisible elements because it seems like we are using the static images as screenshots. So it does not matter if those elements are duplicated on the page.

nobso:
> Replace this line...
https://github.com/garris/BackstopJS/pull/249/files#diff...

We cannot use reduce on ``` scenario.selectors``` because this will not have any of the identical elements. We don't want to ask the developer to mention these similar selectors in there.

nobso:
I can work on making this 'selectorExpansion' as configurable.

garris:
Hi @nobso,  Great to see that you are still working on this when you can!   

I am very excited because the new version 2.0 is feature complete and it is a huge huge improvement!   I am working on the documentation and hope to finish within 2 weeks.  

If this feature is finished in that time I would gladly include it -- but no worries at all if not -- just take your time.

I'll reply to your comments below... Hope all is well!

garris:
>> Replace this line... https://github.com/garris/BackstopJS/pull/249/files#diff...

> We cannot use reduce on scenario.selectors because this will not have any of the identical elements. We don't want to ask the developer to mention these similar selectors in there.


So, below is an example of what selector expansion would do -- and at the end is a psudoCode example of how it could probably be implemented.

When doing a reference run, if you have a DOM layout like this...
```
...
<body>
  <div class="logo">
  <div class="logo">
  <div class="logo">
</body>
```
and a config with this
```
...
"selectors": [
  ".logo"
], 
```
then the expansion script would decorate the DOM like so...
```
...
<body>
  <div class="logo">
  <div class="logo __n_1">
  <div class="logo __n_2">
 </body>
```
and the internal property `expandedSelectors` would look like this...
```
"expandedSelectors": [
  ".logo"
  ".logo.__n_1"
  ".logo.__n_2"
]
```
so below, this "reduce" actually "expands" single entities by doing a DOM query on each one. As a side effect, it also does the DOM decoration.  At least that's the idea.   If it's done in this way, the footprint for this feature is actually very very small.
```
scenario.expandedSelectors = scenario.selectors.reduce(function(acc, selector){ 
  var expandedSelectorArray = [].slice.call(document.querySelectorAll(selector)).map(function(element, i) {
    var nPart = i ? '__n_' + i : '';
    element.classlist.add(nPart); // <== I think this may fail because element is not live -- but this is the idea
    return selector + '.' + nPart;
});
  return acc.concat(expandedSelectorArray);
}, []);
```

I think this is really the heart of the feature.  There are probably a few edge cases (missing selectors across runs?) that will need some sorting out -- but most of it is right here.

nobso:
@garris i've incorporated the suggestion you gave. Actually that's neat and clean! Please let me know if you think that i've miss understood anything. BTW i've tested this feature with single element, multiple similar elements and with ignored elements. Things look fine too me.

garris:
That's great to hear!  I will review as soon as I can.   If possible, could you add a test example to the examples directory.   

Cheers!

nobso:
Yes, sure. I will add that.

nobso:
@garris i've added a simple example folder.

nobso:
screen shot 2016-09-15 at 10 52 31 pm

nobso:
@garris If you have any more comments i can take those up in this weekend as i'm free on Sunday.

garris:
this is super!  Here's comments...

Can you checkout version_2.0 and PR against that?   You will need to copy by hand since file structure is a little different (/test was moved to /example )

Also, you will see that file name generation has changed -- you'll need to do some modifications there.

Maybe use `__n1` as the filename & CSS extension? It needs to be reasonably small and unique since it will occur on reports.  I have high confidence benefits of brevity outway uniqueness.

I think you don't need to exclude the hidden selections from reports -- if someone includes the selector then we should not override their config -- it's not clear to me that this override is going to be obvious in practice.

Thank you!

nobso:
Hi @garris ,

> I think you don't need to exclude the hidden selections from reports

We need this because ..

```html
<div class="container">
    <input class="btn btn-primary" type="submit" value="Submit">
    <input class="btn btn-primary disabled" type="submit" value="Submit">
</div>
```
In this case, there are multiple ```btn-primary``` elements, but If i don't want to test the disabled elements on the page, then i would do something like the below in backstop config,

```json
"removeSelectors": [
     ".container .btn.btn-primary.disabled"
],
"selectors": [
    ".container  .btn.btn-primary"
],
```

This 'disabled' elements should not be part of the expanded selector. That's why i need that.

garris:
Ok.  I understand what the issue is now.   

So, here... https://github.com/garris/BackstopJS/pull/249/commits/6e0496d65841ab88e8a66ecb8cbbb9417dd88583#diff-8854ff0ca17d960be935d14ef5decc3fR152

instead of `display:none`, maybe we should simply remove the selector.  I think "removeSelectors" should actually do what it says.     i.e. `s.parentNode.removeChild( s );`   

That eliminates the need for filtering later on.   Is that more intuitive for you as well?

nobso:
Yes, you are absolutely right. 
But i didn't want to remove as this version is being used by many and it may break.

garris:
Me too.  But I feel like the purpose of a major revision is to add big new features and also to fix errors made before -- even if that breaks _some_ functionality/edge cases.   In this case, I have been regretting that I used the name "removeSelectors" without actually removing the DOM node.  That was a little bit silly.  😝    I will add that note to the documentation -- it probably wont be too bad for most every case.

nobso:
I totally agree that we can add it to this major version

nobso:
@garris Please review https://github.com/garris/BackstopJS/pull/287

garris:
Closing -- moved to new PR.

PR #252, opened 2016-08-15
Added authentification support via cli

svwu:
To enable dynamic authentification for many users on one specific site, we won't need any static username and password in our config, we can now set them via parameters.

### DIFF
--- a/capture/genBitmaps.js
+++ b/capture/genBitmaps.js
@@ -68,10 +68,11 @@ function capturePageSelectors(url,scenarios,viewports,bitmaps_reference,bitmaps_
 
   casper.start();
 
-  if(casper.cli.has('password') && casper.cli.has('username') ) {
-    casper.setHttpAuth(casper.cli.get('password'),casper.cli.get('username'));
-    casper.echo("Use Auth from user " + casper.cli.get('username'));
+  if(casper.cli.options.user && casper.cli.options.password) {
+    console.log('Auth User via CLI: ' + casper.cli.options.user);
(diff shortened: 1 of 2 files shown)


svwu:
@garris 
How do you think about Auth via command line?
I've seen that gulp isn't supported anymore. 
I would rewrite auth via command line for npm, if you're interested. 
We need this feature and waiting for an "official" Solution, which is supported, not to code it everytime new...

Best, 
Sven

garris:
@svwu I apologize for the delay following up -- I have been preoccupied trying to get 2.0 out.   Yes. It is reasonable to add this feature.   And yes, I think it would be great for you to add it.  However, if possible, I would like it added as an onBefore script module.   In this case, you could create a cliArgs object and pass it as a parameter to the onBefore & onReady hook.  Your module can parse for the relevant key values.   The reason for this is because auth can be very complex.  There are many who will need some basic examples they can modify for thier requirements.  I myself am going to publish a gmail login module and a cookie parser module.

If you implement a clean way to do this then it would be easy for any developer to pull arguments into his/her plugin.

What do you think?

svwu:
It sounds great! As my answer comes a bit late, I'm busy atm and guess I'll have time about beginning next month..

svwu:
Hello @garris ,
I just made the change for auth via cli.
Can you please have a look?
Thanks!

svwu:
This is an old version, please have a look to this pullrequest:
https://github.com/garris/BackstopJS/pull/327

Issue #257, opened 2016-08-18
Replace some references and leave the others


SKOLZ:
Hello @garris! First of all, thank you for this awesome tool! I want to start using it in the near future so I was checking out how it works. Although everything is working fine, I would like to know if there's a way to say that a test failure is because of an error on my part or otherwise because I intentionally modified those files, in which case I would want to update the reference images of those specific files.

garris:
@SKOLZ There is a feature enhancement request for this... See #95

SKOLZ:
@garris thank you! I'll check it out

garris:
Closed by #259

Issue #256, opened 2016-08-17
404s in the report


danwellman:
I've got 50 scenarios and 3 viewports, resulting in 150 reference images.

Running the `npm run test` command completes the test run successfully, but when the report page loads up in the browser, it never actually completes.

Either I get 404s for images that I can see in the bitmaps_test/[timestamp]/ folder, or the page gets between 1/4 and 1/2 way through loading all of the images, then crashes (e.g. the 'aw snap' page in Chrome).

I've not seen a successful report yet and have tried numerous times. 

During testing, when I was working with 1 or 2 scenarios, the report page worked perfectly

garris:
Hmm. 404?  I don't think you would see that if it was just a matter of having too many test cases -- but I don't have much context here.   Have you tried the CLI option?  Does that work?

Also, would it work if you were to just run one of the scenarios?

danwellman:
Hello, yes, I'm using the CLI option now, it's working very well with no issues at all. Also, yes, if I run a small selection of scenarios, up to about 20 or so, the HTML page loads fine without any issues

garris:
Ok.  Good to hear.  Our next step is to refactor the web report to handle very large sets.

garris:
Closed by #259

Issue #255, opened 2016-08-17
Generate screenshots for just one scenario


danwellman:
Is it possible to generate new reference screenshots for just a single scenario?

I have 150 reference screenshots for a project, which is a set of UX components. When updating the CSS for a single component, it would be nice to be able to generate new reference screenshots for just the component I worked on, rather than for all components.

garris:
There is an enhancement request for this.  see #95    I would really like to see that one implemented.

garris:
You might also consider breaking your config file up into a few smaller ones and selecting the scenario group you want on the CLI.

danwellman:
I did consider breaking up the config, we do have clear lines of division between the components so it would be relatively easy to do. Thanks

garris:
Closed by #259

Issue #243, opened 2016-08-09
Change location of genConfig


tiagodavi:
Hi Everyone.

I would like to do something like that:

npm run genConfig -- --configPath=./backstop_projects/1
npm run genConfig -- --configPath=./backstop_projects/2

I need to create one backstop_data folder and backstop.json for each project dynamic in my main project. Is it possible?

garris:
Closed by #259

PR #259, opened 2016-08-18
Rework of many parts for speed and CI reporting

onigoetz:
Hello,

first, I'm really sorry for the size of this pull request I know it will be difficult to code review it. We made this changes in a single sprint at our company and couldn't afford to keep everything to separate branches because we needed to push it to our central Jenkins.

## Highlights
- Completely removed gulp, rely only on `fs` to move files around
- Added a modal to show the images in big, with a slider to compare images in place.
- The report doesn't need a server to be seen
- To see the report, the files are copied in the project (previously the images were copied into Backstop)
- Backstop relies on `process.cwd()` to generate its path, not `./node_modules/backstopjs`
- Add a `--filter` option to filter scenarii by `label` in test and reference

## Variants
- You can add "variants" to a scenario, they have the same options as a scenario, but are checked against the main scenario
- no reference is created

## Browser reporting
- Removed resemblejs, use images generated by the CLI reporter. (we have more than 500 tests, running them in the browser just breaks the browser)
- Copy the reporter files to the project.
- The reporter required to disable the CSP rules of our Jenkins server, not longer the case
- Added a way to toggle the summary of scenarii
- Added a modal on click on reference or test image with a comparison slider
- Tweaked the styles a bit
  - Removed custom "circles" to show the number of passed and failed tests : replaced with Bootstrap's "label" because with more than 100 tests, the circle is broken
  - Made a more compact "Report" section : removed `analysisTime` and `dimensionDifference` (if `isSameDimensions` is true)
  - Changed the font stack (there was a typo)

## CLI Reporting
- Reporting is always run in the CLI and results added to an object
- You can easily add multiple new report types with this new object
- junit report was moved to a separate code block to not clutter the main one

## Internal changes
- paths are made from the current working directory, not backstopjs.
- removed the need for an internal hash of the configuration.
- the list of comparisons to create is stored in a special temp file.
- commands no longer have a `before` and `after` this makes the code very complicated to understand and can be replaced by promises inside the main execute
- removed "start" and "stop" commands as the server is not needed anymore.

## Bugfixes / Regressions
- If the tests failed, the return code was still 0
- You could not use a JavaScript configuration anymore
- Errors in Promises were swallowed

## Code Style
- Replaced jshint/jscs with eslint
- Linted with "semistandard"

## Tests
- Run tests on travis on more node versions (0.12, 4, 5, 6)


capture/genBitmaps.js
```diff
@@ -1,251 +1,260 @@
...
+var cwd = fs.workingDirectory;
+var __dirname = null;
+var searchString = "genBitmaps.js";
+require('system').args.forEach(function (arg, i) {
+  var position = arg.length - searchString.length;
+  var lastIndex = arg.indexOf(searchString, position);
+  if (lastIndex !== -1 && lastIndex === position) {
+    __dirname = arg.replace(fs.separator + 'genBitmaps.js', '');
+  }
+});
```
machowski: this looks like duplicated code from 'echoFiles.js', maybe extract it to separate helper file?

capture/genBitmaps.js
```diff
@@ -1,251 +1,260 @@
...
 
-  casper.each(scenarios,function(casper, scenario, scenario_index){
+  casper.each(scenarios, function (casper, scenario) {
+    var referenceId = scenario.label.replace(/\//g, '_');
 
-    casper.each(viewports, function(casper, vp, viewport_index) {
-      this.then(function() {
-        this.viewport(vp.width||vp.viewport.width, vp.height||vp.viewport.height);
-      });
+    processScenario(casper, scenario, referenceId, referenceId, viewports, bitmaps_reference, bitmaps_test, screenshotDateTime, consoleBuffer);
```
garris: `referenceId` is the 3rd and 4th arg?

onigoetz:
I'm not sure that closing the pull request was you goal, or was it ? :)

garris:
see: https://github.com/garris/BackstopJS/tree/version_2_0_onigotez

onigoetz:
Oh great, I didn't see that one.
But how will I be able to push more changes to it ? I have still some changes to make before this PR is completely ready for merge.

And what is your intention with the three branches `version_2_0`,  `version_2_0_onigoetz` and `version_2_0_pradet` ? 

My work being based on what Julien Pradet made in #241 I don't think this should be reviewed separately.

garris:
Ok.  It's merged now.   

This is an epic PR!  This is an overwhelmingly positive direction for BackstopJS.  Thank you for the contribution!

Thanks for clarifying re: #241 -- Your changes should be on 2_0 now.   I will delete the other branches & reopen this PR.

I am intrigued by `scenario.variants`.   

Is it your intention to abandon change detection of the config?

I am already thinking about the various online resources and tutorials that need to be rewritten with this new, much cleaner interface.

garris:
Question:  What is the install process now?

garris:
^^^ Ok.  Got it.  Needs to be a global install.   Carry on...

garris:
This thing is humongous.  Where is the browser report generated now?   The `openReport` command isn't working for me.  Maybe this isn't implemented yet? 

` Error: Command failed: /bin/sh -c open -a "Google Chrome" "/Users/gshipon/Desktop/myCoolProject/html_report/index.html"`

Can we have the configurable location for this default to `./backstop_data/`?

core/util/makeConfig.js
```diff
@@ -28,36 +40,39 @@ function makeConfig (customConfig) {
   // Continuous Integration (CI) report
   config.ci_report = config.customBackstop + '/ci_report';
 
+  // HTML Report
+  config.html_report = config.customBackstop + '/html_report';
```
garris: Can we default to `backstop_data/html_report`?

onigoetz:
It's not mandatory to make a global install, you can add it as a dependency in a project, then then the backstop command is linked in node_modules/.bin/backstop
But if you use an npm script in your project, you refer to it simply by writing "backstop reference"
For the configuration modification, we can re-add it I guess, but it didn't integrate well with the changes we made to the structure.
We made a few modifications that can bypass the need for a configuration change detection:- the name of the images is no longer related to the position in the scenarii list but inferred from the label- if a reference image is missing when executing the tests, we fail the test in question.
For the variants, our use case is quite particular but i guess not entirely unique. We applied Backstop to a component library we created, this library has a reference (html +css) and multiple implementations (React and a java taglib) and we want to ensure that no matter what library we use, we always get the same visual

Get Outlook for iOS

		_____________________________

onigoetz:
The report is genrated in the report command, there is one function for the junit report and one for the browser report, all it does for the browser report is basically take the Reporter class and serialize to json in compare/config.js

Get Outlook for iOS




On Sat, Aug 20, 2016 at 1:52 AM +0200, "Garris" <notifications@github.com> wrote:












This thing is humongous.  Where is the browser report generated now?   The openReport command isn't working for me.  Maybe this isn't implemented yet? 



Error: Command failed: /bin/sh -c open -a "Google Chrome" "/Users/gshipon/Desktop/myCoolProject/html_report/index.html"



Can we have the configurable location for this default to ./backstop_data/?

onigoetz:
And the report opening should work, i tested it on my mac before commiting.I'll see if I can find out why it does this
Get Outlook for iOS




On Sat, Aug 20, 2016 at 1:52 AM +0200, "Garris" <notifications@github.com> wrote:












This thing is humongous.  Where is the browser report generated now?   The openReport command isn't working for me.  Maybe this isn't implemented yet? 



Error: Command failed: /bin/sh -c open -a "Google Chrome" "/Users/gshipon/Desktop/myCoolProject/html_report/index.html"



Can we have the configurable location for this default to ./backstop_data/?

core/util/makeConfig.js
onigoetz: Yes I think we can change that

Get Outlook for iOS




On Sat, Aug 20, 2016 at 1:57 AM +0200, "Garris" <notifications@github.com> wrote:












In core/util/makeConfig.js:
> @@ -28,36 +40,39 @@ function makeConfig (customConfig) {
>    // Continuous Integration (CI) report
>    config.ci_report = config.customBackstop + '/ci_report';
>  
> +  // HTML Report
> +  config.html_report = config.customBackstop + '/html_report';



Can we default to backstop_data/html_report?

garris:
It looks like the `open chrome` command is working fine.  I think the `index.html` isn't getting written.

garris:
Anyway,  Great work!

onigoetz:
Oh and by the way, because we will make a final pass at linting the code, should we use a specific standard ? Something like [standard](http://standardjs.com/) or [semi-standard](https://github.com/Flet/semistandard) ?

garris:
I think semi-standard will create the most happiness among most contributors -- so let's please use that one.

garris:
I just want to double check on this one.  I am not aware of any way to put the `backstop` command on a users path by doing a local install.   

However, your changes to the readme read such that one could successfully run...
```
npm install backstopjs --save-dev
```
 and then from the project directory run...
```
backstop genConfig
```
I understand you *can* do that in the context of a `package.json > script` property -- but this is not the case if one is simply using the command line.  If left as-is, this documentation disconnect would definitely confuse users -- especially new ones.  

To be sure: This (non-PATH-ified) local-install orientation is really not a bad thing given that older implementations (the ones using `npm run <command>` from `node_modules/backstopjs/`) will continue to work as before.  

Please let me know if I missing something here.   If not -- I would propose that the readme is written in the context of a global cli install.  ie. `npm install backstopjs -g`   This would be in-line with other npm based CLI tools like gulp & ember-cli without breaking older implementations.

Again, please let me know if you have any thoughts or info contrary to this.

Best, Garris

onigoetz:
You're right, it's easier to explain to install globally and use the command this way.
With a paragraph explaining how to use the command in npm scripts when installing locally.

garris:
@onigoetz   I've just tested the global and local install.   The browser report is not getting generated in either.   Please take a look at that when you get the chance.

Also, when `browser` is one of the options selected can we run the browser report (with browser self-opening) after testing is done?   I don't know if MacOS and Win need different span commands.

garris:
Looks like the browser report issue is coming from `makeConfig.js`  It doesn't appear that `config.report` is getting set properly there.   That's all I got for now.

garris:
Getting closer.  I just observed everything works (i.e. `makeConfig.js` generates the correct config props) when explicitly setting the config file path.

garris:
Ok. Just made pull request to onigotez/2_0_0 with fixes...

https://github.com/onigoetz/BackstopJS/pull/1


...and it looks like I broke the tests...   😕

In this case -- I think the tests need to be rewritten.

onigoetz:
Hi,
so, I merged your pull request and linted all the code according to eslint-semistandard.

The tests failed on `npm run genConfig` because it couldn't find the config file ... which `genConfig` was supposed to create :D 

So for the moment, I wrapped the loading of the user configuration in a `fs.existsSync` to test if it exists before trying to load it. 
Maybe we can do that in a cleaner way, I don't know.

It seems that the Travis tests don't run because Travis can't open a browser ...
Maybe we should set the `openReport` option to false by default, this will make the tests pass, and allow users to decide by themselves if they want a browser to be opened automatically ?

garris:
Fixes are in https://github.com/onigoetz/BackstopJS/pull/2

I left your fs.exists() condition in the config reading block.  But it just occurred to me that it is actually useful for users to know if there was a missing file error.   If you get to it it would be great to pull that back out.  Otherwise I can do another PR.    Thanks!

onigoetz:
I made a quick check to see if this PR fixes any issues that are already referenced here is the result:

- #164 : Added CLI
- #165 : Done, one part by removing gulp, the other part by cleaning some messages
- #176 : No server anymore
- #205 : this should be fixed 
- #219 : As it is a CLI tool, it's not mandatory to run inside NPM run
- #234 : Using CWD now
- #236 : Removed Gulp, so not needed anymore 
- #237 : Removed Server, so not needed anymore
- #240 : As it is a CLI tool, it's not mandatory to run inside NPM run
- #243 : using CWD, so it's easy to generate a configuration anywhere
- #255 : Added filter for reference creation
- #256 : The pre-made report will prevent crashes
- #269 : Done now
- #273 : no server anymore

Some PR's as well 

- #162 : Server removed
- #199 : Extracted the junit report generation
- #230 : Using a CLI now, probably not needed anymore
- #261 : Failures no longer prevent report opening

And bonus ones

- #137 : Can be closed as #247 was merged
- #257 : Can be closed as duplicated

onigoetz:
By the way, the other people to credit for this work are : @borys-rudenko, @ksushik and @dmitriyilchgmailcom

onigoetz:
Sooo, it seems we're finally there :D 

We made all the changes we wanted and feel confident this pull request is robust and good to merge :D 

Here is a small screenshot of the comparison modal: 

capture d ecran 2016-08-29 a 18 23 48

garris:
Super. I will check this out tonight.  Thank you!

garris:
Awesome -- thank you!   I will be adding some info about the 2.0 version soon.

PR #261, opened 2016-08-19
Enhancement/open report

jcdarwin:
We notice that, if casper tests are run and there are failures, then because `casperChild.on('close'` returns false and doesn't run `gulp.run('report')`.

We're proposing the addition of a new config flag, `openReportIfCasperFailures`, which will ensure that the backstop report is still shown in the browser, even if there are casper failures.

In our case, we're using the xunit file generated by casper to see the casper errors.


garris:
Closed by #259

PR #287, opened 2016-09-19
selector expansion feature is added

nobso:
Adding a selector expansion feature example

### DIFF
--- a/capture/genBitmaps.js
+++ b/capture/genBitmaps.js
@@ -158,7 +158,7 @@ function processScenario (casper, scenario, scenarioOrVariantLabel, scenarioLabe
         scenario.removeSelectors.forEach(function (o, i, a) {
           casper.evaluate(function (o) {
             Array.prototype.forEach.call(document.querySelectorAll(o), function (s, j) {
-              s.style.display = 'none';
+                s.remove();
             });
           }, o);
         });
@@ -169,10 +169,32 @@ function processScenario (casper, scenario, scenarioOrVariantLabel, scenarioLabe
       if (!scenario.hasOwnProperty('selectors')) {
         scenario.selectors = ['body'];
       }
+
+    // if similar selectors exist on the given page, expand and test them too
+    if (scenario.enableSelectorExpansion) {
+        scenario.selectors = scenario.selectors.reduce(function(acc, selector, index) { 
+            var expandedSelectors = casper.evaluate(function(selector, index) {
+                return [].slice.call(document.querySelectorAll(selector)).map(function(element, i) {
+
+                        // by using __n{index}__, we are making sure that this classname does not clash with user defined ones 
+                        var nPart = '__n' + index + '__' + i;
+                        element.classList.add(nPart);
+                        return selector + '.' + nPart;
+                });
+            }, selector, index);
+            return acc.concat(expandedSelectors);
+        }, []);
+    }
+
+
+
       scenario.selectors.forEach(function (o, i, a) {
         var cleanedSelectorName = o.replace(/[^a-z0-9_\-]/gi, ''); // remove anything that's not a letter or a number
         var switchedScenarioOrVariantLabel = (isReference) ? scenarioLabel : scenarioOrVariantLabel;
 
+        if (/__n\d*__\d*/.test(cleanedSelectorName)) {
+            i = cleanedSelectorName.match(/__n(\d*)__\d*/)[1];
+        }
         var fileName = fileNameTemplate
           .replace(/\{configId\}/, configId)
           .replace(/\{scenarioLabel\}/, switchedScenarioOrVariantLabel) // derrived from scenario.label & variant.label
@@ -186,7 +208,7 @@ function processScenario (casper, scenario, scenarioOrVariantLabel, scenarioLabe
           fileName = fileName + '.png';
         }
 
-        var referenceFilePath = bitmapsReferencePath + '/' + fileName;
+        var referenceFilePath = bitmapsReferencePath + '/' + fileName.replace(/__n\d*__\d*/, '');
         var testFilePath = bitmapsTestPath + '/' + screenshotDateTime + '/' + fileName;
 
         var filePath = (isReference) ? referenceFilePath : testFilePath;
--- a/examples/selectorExpansion/backstop.json
+++ b/examples/selectorExpansion/backstop.json
@@ -1,0 +1,43 @@
+{
+    "viewports": [
+        {
+            "name": "phone",
+            "width": 320,
+            "height": 480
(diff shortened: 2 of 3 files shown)


capture/genBitmaps.js
```diff
@@ -169,10 +169,32 @@ function processScenario (casper, scenario, scenarioOrVariantLabel, scenarioLabe
...
       }
+
+    // if similar selectors exist on the given page, expand and test them too
+    if (scenario.enableSelectorExpansion) {
+        scenario.selectors = scenario.selectors.reduce(function(acc, selector, index) { 
+            var expandedSelectors = casper.evaluate(function(selector, index) {
+                return [].slice.call(document.querySelectorAll(selector)).map(function(element, i) {
+
+                        // by using __n{index}__, we are making sure that this classname does not clash with user defined ones 
+                        var nPart = '__n' + index + '__' + i;
```
garris: Can you explain what case we need to guard against here?  I don't see it.

capture/genBitmaps.js
```diff
@@ -169,10 +169,32 @@ function processScenario (casper, scenario, scenarioOrVariantLabel, scenarioLabe
...
+        }, []);
+    }
+
+
+
       scenario.selectors.forEach(function (o, i, a) {
         var cleanedSelectorName = o.replace(/[^a-z0-9_\-]/gi, ''); // remove anything that's not a letter or a number
         var switchedScenarioOrVariantLabel = (isReference) ? scenarioLabel : scenarioOrVariantLabel;
 
+        if (/__n\d*__\d*/.test(cleanedSelectorName)) {
```
garris: Instead of modifying `i` - why not  add a new template key `expandedIndex` and add that to the default template?  We should also make sure this works with the new and currently  undocumented `-- filter=scenarioId` feature.   Will using a filter put our expanded elements out of sync with their index?

capture/genBitmaps.js
```diff
@@ -186,7 +208,7 @@ function processScenario (casper, scenario, scenarioOrVariantLabel, scenarioLabe
           fileName = fileName + '.png';
         }
 
-        var referenceFilePath = bitmapsReferencePath + '/' + fileName;
+        var referenceFilePath = bitmapsReferencePath + '/' + fileName.replace(/__n\d*__\d*/, '');
```
garris: Again, this should probably be done using the template feature.

capture/genBitmaps.js
```diff
@@ -158,7 +158,7 @@ function processScenario (casper, scenario, scenarioOrVariantLabel, scenarioLabe
         scenario.removeSelectors.forEach(function (o, i, a) {
           casper.evaluate(function (o) {
             Array.prototype.forEach.call(document.querySelectorAll(o), function (s, j) {
-              s.style.display = 'none';
+                s.remove();
```
nobso: https://dom.spec.whatwg.org/#dom-childnode-remove - i tested this with Backstop v2. it seems to be working

capture/genBitmaps.js
nobso: This is used to identify the reference image, assume the reference image file name is something like
```
..btn-primary_0_desktop.png
```
After the expansion, it would look like
```
..btn-primary__n0__0_0_desktop.png
..btn-primary__n0__1_0_desktop.png
```

On line number 196, i reset the index to the actual reference file's. So that the script would compare the expanded selectors  against  ```..btn-primary_0_desktop.png```.

capture/genBitmaps.js
nobso: Im not sure, how is that ``` -- filter=scenarioId``` going to work. But if its just used to change the filename like 'configId', 'scenarioLabel' etc, then this selector expansion feature should work just fine.

capture/genBitmaps.js
nobso: Please explain, what do you mean by template feature?

garris: See filenameTemplate string declared earlier in the script.  

I think we should try to not modify name segments named by other features.  This keeps features isolated.  The new filename template should look like this...

{configId}_{scenarioLabel}_{selectorIndex}_{selectorLabel}_{viewportIndex}_{viewportLabel}{expansionIndex};

...and for an element which repeats 3 times expansionIndex should equal '', '__n1', '__n2'.

Does that help?

capture/genBitmaps.js
garris: This will only work if jQuery or similar is included in the target page.  I don't think we are injecting that.  Please review.

capture/genBitmaps.js
nobso: I see, i'll check that

nobso:
@garris sorry i was not available to finish this PR. May i know, why is it closed without any comments, BTW do we have this feature in version 2 already?

garris:
Hello,  very sorry,  i was cleaning up yesterday and deleted to 2.0 branch since we have already done the release. I guess this got closed automatically since it was targeting that deleted  branch.   Please do not be concerned about not completing yet -- I totally understand we are all very busy!   I myself have so much guilt not testing PRs sooner.

I had some time yesterday and patched this to a local branch -- I wanted to see if I could release it but I didn't get a chance to complete testing.  But I really love this feature and want to get it in as soon as possible.   

I will push that branch next time I am on my laptop and we can continue work there -- this feature is really really close.   I'll let you know when I push the new branch you are more than welcome to continue working on it if you get the time.

garris:
And thanks for all your effort!   Hope all is going well.

nobso:
Thanks for the details, @garris  Yes, i see that you are really really busy!

Definitely i would like to work on this and add this feature to our upstream. Yes, please let me know when can i take that up from your new branch. 

Yes, all is well! Thanks for asking :)

garris:
@nobso the feature is in master.  Please take a look when you get a chance.  I tested it today and I thought it behaved well.   I updated the documentation as well.

Let me know if the behaivior is acceptable.   I removed some of the display & filename mutations because I didn't think they were needed.   

The only thing I am not happy with is that the current filenaming doesn't always play nicely when updating `removeSelectors` because the `selector.index` In the file name template is dependent on always having the same number of selectors.   I am still considering going back to a hidden class.   But maybe I am over thinking it.   I do like the current feature footprint is small tho.

Cheers.

nobso:
@garris Thanks!  Sure, I will pull in and check that this evening.

nobso:
@garris upgrading to 2.1.3 on node@6.6.0 does break the tests (tests are running fine in backstopjs v1 on node@6.6.0). For now, it seems to be a relative path issue. I will look into it and give you more details

garris:
Thanks!

Issue #313, opened 2016-10-12
Components failing tests when mismatch percent is near zero.


code-mattclaffey:
Our mismatchThreshold is set as follows: `"misMatchThreshold": 5,` however when the tests run certain tests fail and say `"isSameDimensions": false,`.

Here's an example of some screenshots where the tests _should_ pass but don't seem to.

![prod_test_hillarys_styleguide_30_introbanner_2_desktop](https://cloud.githubusercontent.com/assets/10598938/19303001/bb36ea42-905e-11e6-9fe1-af55ab096095.png)

![prod_test_hillarys_styleguide_30_introbanner_2_desktop 1](https://cloud.githubusercontent.com/assets/10598938/19303013/c434beda-905e-11e6-8fd3-fdaf7936d83b.png)

Here's the output result from the report:

```
Report: {
  "isSameDimensions": false,
  "dimensionDifference": {
    "width": 0,
    "height": -1
  },
  "misMatchPercentage": "0.48"
}
Threshold: 1
```

This happens on around 8-10 tests, intermittently. Are we missing something, or is there a way to turn off the `isSameDimensions` property? Is this expected?

Thanks in advance for your advice.

garris:
I think maybe you should see which element is causing your issue.   Other things to consider: Are you sharing with different machines?  Are there any animations?   Is your CSS math producing half pixel values?   Is content changing from test to test?

There are plans to incorporate a control for this:  but you want to be sure this error isn't exposing a performance issue.  This behavior usually means there is some unknown thing moving your CSS around.

code-mattclaffey:
Thanks ill try that but if my misMatchThreshold is set to 5 shouldn't the test be ignored then because the misMatchPercentage was under that. Is it the isSameDimensions thats throwing it off?

garris:
Currently -- if `isSameDimentions === false` then tests will fail.

Primajin:
I'll try to PR my stuff from #306 soon, then you can at least see in the output that it fails because of different dimensions and not because of missmatch errors.

code-mattclaffey:
Awesome! Please update the issue when you have made the PR.

Thank you :)

Primajin:
See #314 :)

YoungElPaso:
Seems odd that dimensions should be required to be exactly the same.  I guess its fine, but it might be better if there was a fuzziness applied to that no?  A height/width dimension threshold maybe?  I suppose this might just be a matter of opinion.

[...]

PR #314, opened 2016-10-13
more detailed report why a comparison failed.

### LINKED ISSUE #306: show Mismatch in CI output
So back then I added some (I think) useful information to the CI output: when a mismatch error occurs, the output should show how high it is - so I know it's a _real_ regression or its just off by 0.1% and I need to increase threshold ;)

https://github.com/Primajin/BackstopJS/commit/6284fdb5e74b6d71669bc159e37d1da61bc70fff

So it would be great if you guys would add something like this again.

Primajin:
Shows either `(dimensions mismatch)` or `(5% / 3%)` - reads as `five percent mismatch on a three percent threshold`.

Fix #306 , Helps #313

### DIFF
--- a/core/util/compare.js
+++ b/core/util/compare.js
@@ -64,7 +64,12 @@ module.exports = function (config) {
         }
 
         Test.status = 'fail';
-        logger.error('ERROR: ' + pair.label + ' ' + pair.fileName);
+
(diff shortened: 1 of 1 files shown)


garris:
I changed this just a little bit since a test could have one or both failing at the same.   Thanks for the PR!

PR #342, opened 2016-11-03
Adds runner to be used when running locally

klausbayrhammer:
This change should make it possible to that backstop can be used in gulp tasks like this

```
const gulp = require('gulp');
const backstopjs = require('backstopjs');

gulp.task('backstop_reference', () => backstopjs('reference'));
gulp.task('backstop_test', () => backstopjs('test'));
```

In addition I've moved the test scripts from .travis.yml to a npm script - because they are really useful integrationtests for local testing!


garris:
This is great!   I love being able to require into a gulp task or other process! 🌈

PR #354, opened 2016-11-22
fix #329

### LINKED ISSUE #329: BackstopJS Report doesn't fire up after more than 999 tests
Hey @garris,

At 1000 tests and more, the BackstopJS Report doesn't work anymore. I tested it locally at 999 it works and at 1000 it fails.

I don't have any errors. 

This what I get in the CMD : 

```
CasperJS:  Ready event received.
CasperJS:  Current location is http://localhost:4000/documentation/3.3/composantes.html?backstop=tru
e#graphiques
CasperJS:  Capturing screenshots for desktop (1280x1014)
CasperJS:  Ready event received.
CasperJS:  Current location is http://localhost:4000/documentation/3.3/composantes.html?backstop=tru
e&touch=true#graphiques
CasperJS:  Capturing screenshots for phone (320x480)
CasperJS:  Ready event received.
CasperJS:  Current location is http://localhost:4000/documentation/3.3/composantes.html?backstop=tru
e&touch=true#graphiques
CasperJS:  Capturing screenshots for tablet_v (768x1024)
CasperJS:  Ready event received.
CasperJS:  Current location is http://localhost:4000/documentation/3.3/composantes.html?backstop=tru
e&touch=true#graphiques
CasperJS:  Capturing screenshots for tablet_h (1024x768)
CasperJS:  Ready event received.
CasperJS:  Current location is http://localhost:4000/documentation/3.3/composantes.html?backstop=tru
e&touch=true#graphiques
CasperJS:  Capturing screenshots for desktop (1280x1014)
CasperJS:  Comparison config file updated.

Bitmap file generation completed.
COMMAND | Executing core for `report`

C:\Projets\N3\fwd-webapp-overlay-skin>

```

Thanks!

paradizex:
https://github.com/garris/BackstopJS/issues/329

### DIFF
--- a/core/util/compare.js
+++ b/core/util/compare.js
@@ -7,7 +7,7 @@ var streamToPromise = require('./streamToPromise');
 var Reporter = require('./Reporter');
 var logger = require('./logger')('compare');
 
-var MAX_COMPARISONS_CONCURRENCY = 20;
+var MAX_COMPARISONS_CONCURRENCY = 1;
 
 function storeFailedDiffImage (testPath, data) {
   var failedDiffFilename = getFailedDiffFilename(testPath);
@@ -62,6 +62,8 @@ module.exports = function (config) {
         if (data.isSameDimensions && data.misMatchPercentage <= pair.misMatchThreshold) {
           Test.status = 'pass';
           logger.success('OK: ' + pair.label + ' ' + pair.fileName);
+          data=null;
+          pair.diff.getDiffImage=null;
 
           return pair;
         }
@@ -71,6 +73,8 @@ module.exports = function (config) {
 
         return storeFailedDiffImage(testPath, data).then(function (compare) {
           pair.diffImage = compare;
+          data=null;
+          pair.diff.getDiffImage=null;
 
           return pair;
         });


paradizex:
I didn't get time to do through testing but I can provide you with some statistics.

Before this fix the memory usage for 6 url's with 2 viewports (1024x768 and 1920x1080) used to be approx. 2.5GB 
After this fix memory consumption now varies between 500 and 800MB.

I wont be able to post my config's as it points to our local intranet sites which is not accessible from outside the network.

Here are some of the links which I referred to come up with this fix
1.  https://www.toptal.com/javascript/10-most-common-javascript-mistakes#common-mistake-3-creating-memory-leaks
2.  http://info.meteor.com/blog/an-interesting-kind-of-javascript-memory-leak

klausbayrhammer:
Could we make `MAX_COMPARISONS_CONCURRENCY` configurable in the config (since I think that's a cool feature)? If we set it to 1 we could remove the whole concurrency block

paradizex:
That can be a cool feature, but from some of the resources that I have read says that nodejs is a single threaded and that promises do not leave their resources until all of the promises in queue are resolved. So, I assume that, even if we increase concurrency, it will still use single thread and at the same time keep on consuming more memory resource. I have not thoroughly tested my above assumptions but I think that there might be some minor gains in processing of the promises but with higher memory usage.

One other thing I noticed with my fix and setting concurrency as 1, is that after every promise gets resolved, nodejs runs its Garbage Collection routine which in turn increases the total execution time. So, option to configure `MAX_COMPARISONS_CONCURRENCY ` can be useful then. So, in case some one has enough memory then they can increase this value so that GC is deferred till that set of promise queue is resolved, an may get performance gain as a result.

garris:
This sounds awesome!  I am traveling in Asia until next week.   Can't wait to test it.

Also +1 on the config.concurrency parameter!

garris:
Thanks for this fix!!!

I also implemented a config parameter `asyncCompareLimit` to adjust concurrent comparisons.

https://github.com/garris/BackstopJS/releases/tag/v2.3.4

Here is memory usage for various values of `asyncCompareLimit `:
```
1      126mb
10    165mb
100  507mb
```

This will make users very happy -- Cheers!  🏆

paradizex:
That's awesome. 

I respect all the dedication and hard work you have been committing to this project.

Cheers and keep up the good work.

PR #362, opened 2016-12-05
Allow expecting a certain number of tests

### LINKED ISSUE #360: How to specify the expected number of screenshots?
I would like to be able to specify the number of screenshot comparisons that I am expecting before running the test. My use case: I have a working backstop setup, reference screenshots, reports, all that good stuff. Unfortunately, from time to time my code breaks in a way that makes my selectors return 0 results. In this case, running `backstop test` does nothing: there is zero comparisons made and zero comparisons failed, so, success.

I would like to have some help on handling this issue. I would like to set the number of expected screenshots for every scenario so that doing less comparisons than expected would result in test failure, but I don't know how to implement it. Another possible solution would be to decouple screenshot taking from screenshot comparison (possibly related to #293) so that I could compare externally with other test runners.

Thanks!

n1313:
Fixes #360

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ BackstopJS automates visual regression testing of your responsive web UI by comp
 - Use as a standalone global app, a standalone local npm script or import right into your node app.
 - Also plays nice with source control -- share your gold master with your team.
 
-## Upgrade to 2.0 for enhanced speed and new features! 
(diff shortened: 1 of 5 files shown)


klausbayrhammer:
I think we need to do the same check in the CLI-reporter (core/util/compare.js)

PR #366, opened 2016-12-13
--testReportFileName parameter is added for customized xunit filenames

mirzazeyrek:
While using label filters existing xunit reports was being overwritten. This PR solves this issue by using custom report filenames via CLI parameter:

`backstop test --filter="label-1" --testReportFileName=label-1`

or

`backstop test --testReportFileName=custom`


mirzazeyrek:
merged with https://github.com/garris/BackstopJS/pull/380

Issue #363, opened 2016-12-06
Avoid writing to node_modules/backstopjs/


gilligan:
I noticed that backstopjs is creating temporary files such as `config.json` in `node_modules/backstopjs/config.json`. I would like to suggest to use `os.tmpdir` instead.

My case might be special in that my `node_modules` folder is actually read-only but I really think that `node_modules` should only ever be modified by `npm` and not by random packages storing temporary data there. It does not seem like an idiomatic behavior to me either - I haven't had this issue with any other package so far.

Would there be any objections against storing temporary data in directory below `os.tmpdir()` ?

garris:
No objections here.   I won't be available to review till at least January -- that would the only issue potentially on your end.   But sounds like something others would like also.   Cheers.

mirzazeyrek:
I strongly agree with this.  

Another important issue is that while trying to execute concurrent backstop tasks, wrong config.js files are being used because right config.js file would be overwritten... 

That config file name should be unique for avoiding such conflicts. May be sha1 similar function can be used for creating unique ids. Any suggestions ?

garris:
+1 ^^^

svwu:
That would be awesome, cause I don't know how parallel tests on the same machine would behaviour

mirzazeyrek:
old PR closed. New PR merged with custom xunit filename PR:

https://github.com/garris/BackstopJS/pull/380

PR #368, opened 2016-12-19
temp folder for hashed captureConfigFile.

mirzazeyrek:
Avoid writing to node_modules/backstopjs/

Issue described in here:

https://github.com/garris/BackstopJS/issues/363

few eslint warnings are fixed also.

### DIFF
--- a/core/util/extendConfig.js
+++ b/core/util/extendConfig.js
@@ -1,7 +1,10 @@
 var path = require('path');
 var temp = require('temp');
+var fs = require('fs');
+var hash = require('object-hash');
+const tmpdir = require('os').tmpdir();
 
-function extendConfig(config, userConfig) {
+function extendConfig (config, userConfig) {
   bitmapPaths(config, userConfig);
   ci(config, userConfig);
   htmlReport(config, userConfig);
@@ -18,7 +21,7 @@ function extendConfig(config, userConfig) {
   return config;
 }
 
-function bitmapPaths(config, userConfig) {
+function bitmapPaths (config, userConfig) {
   config.bitmaps_reference = path.join(config.projectPath, 'backstop_data', 'bitmaps_reference');
   config.bitmaps_test = path.join(config.projectPath, 'backstop_data', 'bitmaps_test');
   if (userConfig.paths) {
@@ -27,7 +30,7 @@ function bitmapPaths(config, userConfig) {
   }
 }
 
-function ci(config, userConfig) {
+function ci (config, userConfig) {
   config.ci_report = path.join(config.projectPath, 'backstop_data', 'ci_report');
   if (userConfig.paths) {
     config.ci_report = userConfig.paths.ci_report || config.ci_report;
@@ -47,7 +50,7 @@ function ci(config, userConfig) {
   }
 }
 
-function htmlReport(config, userConfig) {
+function htmlReport (config, userConfig) {
   config.html_report = path.join(config.projectPath, 'backstop_data', 'html_report');
   config.openReport = userConfig.openReport || true;
 
@@ -59,17 +62,25 @@ function htmlReport(config, userConfig) {
   config.compareReportURL = path.join(config.html_report, 'index.html');
 }
 
-function comparePaths(config) {
+function comparePaths (config) {
   config.comparePath = path.join(config.backstop, 'compare');
   config.tempCompareConfigFileName = temp.path({suffix: '.json'});
 }
 
-function captureConfigPaths(config) {
-  config.captureConfigFileName = path.join(config.backstop, 'capture', 'config.json');
+function captureConfigPaths (config) {
+  var captureDir = path.join(tmpdir, 'capture');
+  if (!fs.existsSync(captureDir)) {
+    fs.mkdirSync(captureDir);
+  }
+  var tempCompareConfigFileName = config.tempCompareConfigFileName;
+  config.tempCompareConfigFileName = '';
+  var configHash = hash(config);
+  config.tempCompareConfigFileName = tempCompareConfigFileName;
+  config.captureConfigFileName = path.join(tmpdir, 'capture', configHash + '.json');
   config.captureConfigFileNameDefault = path.join(config.backstop, 'capture', 'config.default.json');
 }
 
-function casper(config, userConfig) {
+function casper (config, userConfig) {
   config.casper_scripts = path.join(config.projectPath, 'backstop_data', 'casper_scripts');
   config.casper_scripts_default = path.join(config.backstop, 'capture', 'casper_scripts');
 
--- a/core/util/extendConfig.js
+++ b/core/util/extendConfig.js
@@ -72,10 +72,7 @@ function captureConfigPaths (config) {
   if (!fs.existsSync(captureDir)) {
     fs.mkdirSync(captureDir);
   }
-  var tempCompareConfigFileName = config.tempCompareConfigFileName;
-  config.tempCompareConfigFileName = '';
   var configHash = hash(config);
-  config.tempCompareConfigFileName = tempCompareConfigFileName;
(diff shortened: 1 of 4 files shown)


garris:
This looks great!  I think this would enable multiple instances to run in parallel -- yes?

garris:
@klausbayrhammer, @jtuds, @svwu,

Please take a look if anyone has a few minutes.   

Cheers.

Garris

mirzazeyrek:
Yes actually main benefit is being able to run multiple instances.

core/util/extendConfig.js
```diff
@@ -59,17 +62,25 @@ function htmlReport(config, userConfig) {
...
-  config.captureConfigFileName = path.join(config.backstop, 'capture', 'config.json');
+function captureConfigPaths (config) {
+  var captureDir = path.join(tmpdir, 'capture');
+  if (!fs.existsSync(captureDir)) {
+    fs.mkdirSync(captureDir);
+  }
+  var tempCompareConfigFileName = config.tempCompareConfigFileName;
+  config.tempCompareConfigFileName = '';
+  var configHash = hash(config);
+  config.tempCompareConfigFileName = tempCompareConfigFileName;
```
svwu: From the line 75 to 78, I don't know why you save config.tempCompareConfigFileName in a var, set the config.tempCompareConfigFileName to '' and set it back to the var?

mirzazeyrek: makes sense. I'll update it.

svwu:
The idea behind parallel processes doesn't work, cause I'll get same hashes on two different processes. Maybe you add a timestamp to the hash?
And also the new config remains in the temp folder. It would be better if it will be deleted afterwards.

mirzazeyrek:
@svwu no, it works. this is an expected behavior. if hash is same it means file is the same. why need create different hash ? 

yes there is no delete afterwards. it can be improved. any contribution is welcome.

core/util/extendConfig.js
mirzazeyrek: Okey, I have checked this and remembered why I do like this.

This is an intended behavior for preventing different hashes because of a dynamic temp value. 

The hash value only needs to be change when you make a change in your backstop.json file. 

If we do this as your suggestion every time we are going to get different hash name and it will create a lot of tmp files (even backstop.json file is same) for each execution.

svwu:
I've updated my comment, some minutes before you submitted your comment. 
Yes, you're definatly right, that it works. But It's not working if you run the same config. 
But the situation could be, that we have a onBeforeScript, multiple viewports and multiple scenarios. If the test lasts longer but another user wants to run the same config on the same server, it wouldn't work.

So maybe we simple add a timestamp to the value which will be hashed?

mirzazeyrek:
Could you please share your files and commands which you were using for test. I can't reproduce the error. @svwu

svwu:
backstop reference --user="test" --password="test"

of two different cmds

svwu:
The Error I get (in Windows):
`COMMAND | Error: Cannot find module 'C:\Users\sven\AppData\Local\Temp\117018-11436-ayglre.json'
              at Function.Module._resolveFilename (module.js:336:15)
              at Function.Module._load (module.js:278:25)
              at Module.require (module.js:365:17)
              at require (module.js:384:17)
              at module.exports (C:\Users\sven\AppData\Roaming\npm\node_modules\backstopjs\core\util\compare.js:48:23)
              at Object.module.exports.execute (C:\Users\sven\AppData\Roaming\npm\node_modules\backstopjs\core\command\report.js:97:12)
              at Object.execute [as report] (C:\Users\sven\AppData\Roaming\npm\node_modules\backstopjs\core\command\index.js:49:49)
              at execute (C:\Users\sven\AppData\Roaming\npm\node_modules\backstopjs\core\command\index.js:94:31)
              at C:\Users\sven\AppData\Roaming\npm\node_modules\backstopjs\core\command\test.js:9:14
              at runMicrotasksCallback (node.js:337:7)
              at process._tickCallback (node.js:355:11)`

I check Ubuntu, too.

core/util/extendConfig.js
svwu: It's now my own opinion speaking (you may see it in a different way):
but why do we need the temporary config after the test? Cause we allready got the config file. It's not a great deal to generate a temporary file again. 
It's just a waste of space, if we're about to run multiple tests as we got many different settings, we get a huge load of files. So why we won't keep it really temporary, as the filename intends and delete after the execution? With the hashed value the timestamp and the config, we guarantee that we only use the temporary file as long as the test will be executed and it won't remains dead on our temp directory. 
An automatically garbage collection of our temporary testfiles would be nice.

mirzazeyrek: I am able to reproduce the error on my windows machine. I'll take a look tomorrow. Thank you for noticing this. I couldn't thought about a scenario for execution of same test at the same time. (still I can't :) )

garris:
I think the use case for running the same config twice is when using the `--filter` argument.
```
backstop test --filter=^loginFlow$ //runs for 5 mins

// 2 mins later

backstop test --filter=^userWidgets
```

Using timestamp is just a little fragile.  Best case is to maybe to check for an existing hash and increment an index.

mirzazeyrek:
@garris different filters are working without an issue for same backstop. Json file. Actually we are using it that way in our ci process. Issue is for exact same test case on Windows as far as I can confirm so far.

svwu:
I can't guarantee that I'm the only one who executes tests on our server or maybe the same test. As I build a webinterface to make it won't affect other people running on the same server tests. This is my use case starting these tests during development before any commit or for users who just test without doing any code changes, to check that we won't have any regression.

garris:
@mirzazeyrek  could you add mocha tests for this feature to the tests directory?  Please include any other tests which demonstrate the behavior.  Looks like we still need to track down why some cases fail.  Looking forward to it -- this is a cool feature!

mirzazeyrek:
I have tested on windows and ubuntu. Can we confirm that issue is resolved with new commit ? @svwu  @garris

mirzazeyrek:
btw, I have noticed backstopjs is already creating temp files and not deleting it. It can be improved with a clean up task... 

image

mirzazeyrek:
@svwu were you able to check it ? please let me know.

garris:
Thank you very much guys for working on this cool feature.

@mirzazeyrek after we get a good review, we'll put it on a new canary branch and you can start using that in your organization.   I'll notify others at LinkedIn to update as well.  After a bit we'll promote it to master.

garris:
@mirzazeyrek I know you've been very patient while waiting for a full review.    I just created a beta branch -- please feel free to go ahead and create a PR to that branch.  I will pull that in and push it out to NPM @beta and we can start asking others to test.

mirzazeyrek:
Thank you @garris . I'll create that PR and send you. About using this feature in our CI process, Unfortunately it's not possible without two other PR's that I've sent.

We are using custom xunit file names, which is being passed from cli. That way it won't be overwritten by every different filter executing at the same time.

And also JPEG support due to huge number of tests, png images are contains huge space.

garris:
The xunit file names looks good.  Please add that also.    

I am confused a little by the JPEG support PR.  The logic for parsing the extension with a default is a little clunky.   But mainly Indont know why the two formats require two different write methods.

Please let me know about that  thanks.

svwu:
Sorry for the delay, had personal issues to sort.
I reviewed the new change and it works fine, I get differente hashes for the two processes and they won't throw any error, yet.

mirzazeyrek:
I have combined two PR's into one for beta branch. Please let me know if anything else is needed.

About the jpeg output support we provided a different method for JPEG because these two formats uses different libraries for encode/decode in `node-resamble-js`.

mirzazeyrek:
merged with https://github.com/garris/BackstopJS/pull/380

PR #377, opened 2017-01-21
JS object can be passed as config

skip405:
We can pass a JS object directly as a config. Please see the updated `README.md` (`You can pass a config object directly` section) for examples.


garris:
This is GREAT!  Will review this week.

garris:
+1 on the config.js example.

skip405:
Hi guys. Thanks for accepting this PR. How shall I give an example? A more detailed `README.md` description? I have it there? but it's kind of small, yes

garris:
@skip405 It would be very very cool if you could add a super simple project in the examples directory.  E.g. `backstopjs/node_integration/` and you could have an index that maybe prints `hello world` after checking google.com or something.   I think something like that would really help other developers.   Would be great to see if it's not too big of an ask.   

Either way -- thanks for the contribution and please do a pull request to the `beta` branch and I will get that pushed to npm@beta asap.

Thanks!

stevecfischer:
I have a suggestion but it comes with a caveat. Since there already is the `capture/config.default.json` create a `capture/config.default.js` and update method to copy the js just like the json file.

The caveat is I'm not particularly fond of those "defaults" automatically being created as a result of `genConfig()`. What I'm trying to say is `genConfig()` is necessary to create the folder structure, but I don't necessarily want these default files when I create new projects. Maybe add a flag to copy examples or some other way to decouple what is required from `genConfig()` vs what is optional.

garris:
@stevecfischer 
What I'd prefer is to just add an example e.g. `backstopjs/examples/js_based_config` and just show a simple example.  It could even just import a backstop.json file and mutate one of the props. We can point to the example in the readme.

And in any case, I totally agree with you on the second bit.  Sometimes I just want to generate the scaffolding but not clobber the existing config.   I was thinking it would be nice for backstop to simply check for a backstop.[js|json] and prompt to overwrite (with the default being no).   I would pull that in right away.

garris:
@skip405 Please disregard my comment to @stevecfischer.   It probably belongs in it's own thread.   

@stevecfischer  I like this line of thought but you should probably open a new thread if you want to pursue this very good suggestion.   thanks!

stevecfischer:
@garris - yea your right. I'll run with the idea.

skip405:
Closing this PR in favour of the [new one into another branch](https://github.com/garris/BackstopJS/pull/386).

PR #383, opened 2017-01-25
added jpg support.

mirzazeyrek:
For saving bandwidth and space especially with huge test cases and CI tool integrations. 

screenshot_80

added suggested changes from @garris

### DIFF
--- a/capture/genBitmaps.js
+++ b/capture/genBitmaps.js
@@ -15,6 +15,7 @@ if (!config.paths) {
   config.paths = {};
 }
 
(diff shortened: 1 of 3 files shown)


garris:
Will push this to npm today.

marc0n3:
Is jpg compression level selectable somewhere?

mirzazeyrek:
Hi @marc0n3 ,

Yes but not via config file unfortunately. There are some changes since my jpeg PR but changing 85 in getDiffImageAsJPEG method should work:

https://github.com/garris/BackstopJS/blob/master/core/util/compare/store-failed-diff.js#L20

PR #386, opened 2017-01-27
JS object can be passed as config

skip405:
Hi guys,

doing a new PR into the `beta` branch. Please take a look at the `nodeIntegration` example I have there. It is about having one config to test several similar projects + it passes some parameters to the config file (in this case - `scenarios`, `id`, `paths`) and then imports it. Are we OK with this example?

Please note that since I'm actually installing `backstopjs` via npm in that example, it doesn't have the updated version yet :) So to test my example, you'll have to update `createBitmaps.js` and `makeConfig.js` in the `node_modules` folder of my example with the code from this PR. Then it should work. Sorry if it causes troubles with you.

I'm going to close my previous PR now.


garris:
@skip405 -- I resolved a conflict, merged, and pushed to backstopjs@beta. 

thanks @stevecfischer for the review!

skip405:
@garris Fantastic :) Everything's OK with my example, I presume?

garris:
I only had time to skim it.  It is a little more complex than I would have shown.  _The thinking here is that the added complexity (files etc...) makes it less obvious what feature you are demonstrating -- that said, a more invested user may potentially learn a little more from all that added complexity._  

With documentation it can be hard to get the right amount of detail -- as feature creators there is often so much we want to show, the subtleties etc... At the same time users who have resorted to looking at the example are often skimming, looking for a specific hint which will get them over the particular learning curve they are in.  So sometimes these two intentions things can be at odds.

In any case -- the feature is really smart and it's awesome you took the time to add the example -- so lets just go with it!

Cheers!

skip405:
OK, I see. Yes, I quite agree with you that it's hard to find this balance. I first tried a simpler example, but it ended up being super simple, like - using your config from the JSON file directly as a JS object and that's it )) But thought it would not have been much of a NodeJS integration then, that indeed resulted in some complexity. Hope users will understand it. Anyways, it was a pleasure to contribute. Cheers!

PR #395, opened 2017-02-20
Various bugfixes

onigoetz:
Hi @garris,

I tried to update backstop and got a few surprises.

- When an error was thrown in `compareImage` not a single report was shown. (just stopped the process with an exit code of 0)
- Variants were not compared to the base reference image, but to a reference image that was never generated

So I made the following bugfixes for more robust error handling : 
- Centrally catch and log the unhandled errors
- use `allSettled` for report handling, that gives a change for a report to fail and give feedback on that error
- use `p-map` instead of `bluebird.map` because bluebird swallowed errors
- return errors in `compareImage` as `resolve(new Error(...)` so that we can accurately report all errors in image comparisons.

Other changes
- fix a typo in `core/command/report` for `testReportFileName`
- fix the way the filenames are generated when generating the bitmaps.

### DIFF
--- a/capture/genBitmaps.js
+++ b/capture/genBitmaps.js
@@ -94,6 +94,25 @@ function capturePageSelectors (scenarios, viewports, bitmapsReferencePath, bitma
   });// end casper.each scenario
 }
 
+function getFilename(scenarioIndex, scenarioLabel, selectorIndex, selectorLabel, viewportIndex, viewportLabel) {
+  var fileName = fileNameTemplate
+    .replace(/\{configId\}/, configId)
+    .replace(/\{scenarioIndex\}/, scenarioIndex)
+    .replace(/\{scenarioLabel\}/, scenarioLabel)
+    .replace(/\{selectorIndex\}/, selectorIndex)
+    .replace(/\{selectorLabel\}/, selectorLabel)
+    .replace(/\{viewportIndex\}/, viewportIndex)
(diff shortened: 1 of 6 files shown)


garris:
@onigoetz thank you so much for the fixes!   I will review this week (I will hopefully have some time Friday)

garris:
Sorry,  I was doing some house cleaning and blew away the canary branch which accidentally closed this PR.   I will try to generate a patch from this commit and reinstate canary.

garris:
Boom.  Ok.  there we go!

Looks like the node v1.x test may have failed.   I am going to push this out anyway.

garris:
ok. I am going to go ahead and push this to @canary.  There is a broken test on `node v0.12.18`.   If you get a chance it would be great if you take a look and see if anything jumps out at you.  I'll try to address it if I get some time.   Thank you so much for the bug fixes!

garris:
Travis log here... https://travis-ci.org/garris/BackstopJS/jobs/204804551

garris:
Now available via `npm install backstopjs@canary -g`

onigoetz:
Hi, sure I can have a look.

The support fort node 0.10 and 0.12 has ended recently, do you really want ton continue supporting it ?

garris:
I know.  I really don't think it's necessary to support.  I have no problem dropping that test.

garris:
FYI: I just removed the v0.12 test.

PR #398, opened 2017-02-22
fixes config.openReport flag

pgn-vole:
`config.openReport` flag was not working as extend config was always setting it to true.
It now uses the value provided in the config file.


pgn-vole:
@garris sure. I made the change.

garris:
Great!  Thank you!

pgn-vole:
@garris do you mind making a patch release for this change? We would like to start using it :)

garris:
Ok.  It's published.  Go ahead and update with `npm install backstopjs -g`

PR #404, opened 2017-02-28
New updated workflow feature: backstop approve

garris:
`backstop approve` will copy bitmaps from your latest test into your reference directory.  Subsequent tests will be compared against your updated reference files.

Provides a straightforward way to approve changes.

I could also see adding `--filter=<scenario.label || selector>` and `-d` (dry run) arguments to enable approving a more narrow selection.

### DIFF
--- a/core/command/approve.js
+++ b/core/command/approve.js
@@ -1,0 +1,25 @@
+var fs = require('../util/fs');
+var path = require('path');
+var map = require('bluebird').map;
+
+var FAILED_DIFF_RE = /^failed_diff/
+
+// This task will copy ALL test bitmap files (from the most recent test directory) to the reference directory overwritting any exisiting files.
+module.exports = {
+  execute: function (config) {
+    // TODO:  IF Exists config.bitmaps_test  &&  list.length > 0n  (otherwise throw)
+    console.log('Copying bitmaps from ' + config.bitmaps_test + ' to ' + config.bitmaps_reference);
+    return fs.readdir(config.bitmaps_test,(err, list) => {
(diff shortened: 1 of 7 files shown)


core/command/approve.js
```diff
@@ -0,0 +1,26 @@
...
+// This task will copy ALL test bitmap files (from the most recent test directory) to the reference directory overwritting any exisiting files.
+module.exports = {
+  execute: function (config) {
+    // TODO:  IF Exists config.bitmaps_test  &&  list.length > 0n  (otherwise throw)
+    console.log('Copying from ' + config.bitmaps_test + ' to ' + config.bitmaps_reference + '.');
+    return fs.readdir(config.bitmaps_test,(err, list) => {
+      var src = path.join(config.bitmaps_test, list[list.length-1]);
+      return fs.readdir(src, (err, files) => {
+        console.log('The following files will be promoted to reference...');
+        return map(files, (file) => {
```
klausbayrhammer: Probably using filter.map would be more expressive:
```
return files.filter(file => FAILED_DIFF_RE.test(file))
        .map(files, file => {
           console.log('> ', file);
           return fs.copy(path.join(src, file), path.join(config.bitmaps_reference, file));
         })
```

garris: Yes that is totally the pattern I wanted -- but `map` here is actually `p-map` (line 3) which wraps a generic map fn in a Promise.all().  So, not chainable here etc.

garris:
@klausbayrhammer  Awesome!  me too -- I really like this feature.  
Thank so much for the review!

PR #405, opened 2017-03-01
fixes 0 not being recognized as misMatchThreshold value

robertmaier:
Based on https://github.com/garris/BackstopJS/issues/127 this commit adds the possibility to set the threshold to exactly 0.

At the moment this is impossible because of the way the parameter is read from the scenario or config.


garris:
@robertmaier @klausbayrhammer  Thank you for the improvement and the review!

PR #409, opened 2017-03-07
WIP: Feature/webdriver support

fincha:
improved webdriver io support, still have some open tasks, to fully support the api.


garris:
This is a great feature concept!    I haven't had a chance to review any of it.   I assume you bypass the Casper layer to make it work?

README.md
```diff
@@ -197,7 +197,7 @@ As a new user setting up tests for your project, you will be primarily concerned
 - **`scenarios[n].label`** – Required. Used for screenshot naming.
 - **`scenarios[n].url`** – Required. Tells BackstopJS what endpoint/document you want to test.  This can be an absolute URL or local to your current working directory.
 - **`scenarios[n].selectors`** – An array of CSS selector strings enabling you specify what part of your DOM you want to test.  The default value is `document`, which will attempt to capture your entire layout.
-
+- **`engine`** - by default is "phantomjs" to use webdriver set it to "chrome"
```
garris: Wouldn't the option be 'webdriver' or 'webdriver:browserArg' here?

capture/genBitmapsWebdriver.js
```diff
@@ -0,0 +1,167 @@
...
+    config.viewports.forEach(function (vp, viewportIndex) {
+        
+        console.log('Capturing screenshots for ' + makeSafe(vp.name) + ' (' + (vp.width || vp.viewport.width) + 'x' + (vp.height || vp.viewport.height) + ')', 'info');
+        
+        let l = scenario.selectors.length;
+        let counter = 0;
+        
+        scenario.selectors.forEach(function (selector, i) {
+            let cleanedSelectorName = selector.replace(/[^a-z0-9_\-]/gi, ''); // remove anything that's not a letter or a number
+
```
garris: Functions like these should probably be broken out into separate util files with tests to make this refactor cleaner.

README.md
fincha: sure, makes sence.

fincha:
Is there a pull request from @beta or how you mean this? Right now I am a bit off the master, you did some changes I have to rebase.

uglow:
@fincha @garris Will this PR lead to consistency in font-rendering across OSes when using Chrome via WebDriver?

fincha:
@uglow I have tested this on ubuntu, this worked fine. No idea about windows.

chrishelgert:
@fincha tried it on windows, but the selenium/webdriver didn´t start -  i only saw
```
> visual-regression@1.0.0 test D:\workspace\visual-regression
> backstop test

BackstopJS loading config:  D:\workspace\visual-regression\backstop.json

COMMAND | Executing core for `test`
```

and then the process ended

bitjson:
I'd love to see this feature completed – is anyone working on it right now?

garris:
I'd love to see this too.   I don't think it's being worked on now.   The codebase has changed a bit since this PR -- if someone can rebase against master and validate that it still works on osx we should cut a branch for it.

fincha:
I had to stop working on this for now, the company I was working with chosen different way... so sorry guys, maybe some one can pick this up.

patricknelson:
Just wanted chime in support for this. I’m investigating alternatives to Wraith and I’m very happy to see that not only does BackstopJS have first-class support for Docker, so it’s also awesome to see that there’s some desire to move toward leveraging Selenium/Webdriver for even more flexibility.

I’m not familiar with the internals of this project, but superficially, it looks like it interfaces directly with `puppeteer` to handle running Chrome directly on the local machine via an `engine` configuration. Maybe we could implement Firefox, IE, Edge and support for any browser Selenium supports by specifying `selenium` as an engine (first layer), then the next layer would be to configure it with some sane defaults, e.g. also point to a Selenium sever running on `localhost:4444`. Now since we’ve abstracted this to the network layer, we can then intuitively support Docker, `docker-compose` or even Kubernetes.

Again, only looked at this for an hour or so (burning midnight oil here) but I’m thinking BackstopJS/Selenium/Docker would be a really kickass combination! 😎

PR #424, opened 2017-03-20
fixes headings in README

sturobson:
There's a few H2s, H3s, and H4s that aren't being rendered correctly as there's no space betwixt the `#` and the heading. I've gone through and added spaces where needed


sturobson:
I see @pkra has already raised a PR for this (#43) ... closing this one :)

PR #432, opened 2017-04-11
Set compare html charset to UTF-8

Enchiridion:
I was trying to use symbols in my scenario labels, but they were coming out garbled because the browser wasn't interpreting it as UTF-8.  This adds a meta tag to make the compare report UTF-8 so they render correctly.


garris:
Looks great.  Thanks for the contribution!

Issue #421, opened 2017-03-16
Clean cookies and session each scenario


mantovanig:
Hi,

is possible clean cookies and session for each scenario?
For example if i want repeat the login flow in each scenario.

I've tried to do this in onBefore.js:

```
module.exports = function (casper, scenario, vp) {
  
  casper.page.clearCookies();

  console.log('onBefore.js has run for '+ vp.name + '.');
};
```

But doesn't work

Thanks

garris:
Not clearing cookies is a bug. 

Are you able to set the cookie to another value?

mantovanig:
Ok.

And there is another solution for have a new session in each scenario?
Or i must lunch backstop('test') with one scenario at once?

Thanks

garris:
Yes,  I think the work-around is to use one config per `onBefore` script.   Please let me know if that works for you.

garris:
We should probably explicitly declare that as a known issue.

garris:
closed by #442

PR #442, opened 2017-05-11
Add cookies to bitmaps right before each scenario has run

shanemcgraw:
This will set each cookie to its relevant scenario right before its run, rather than setting all of the cookies at the start of the casper instance. The upshot to this is that we'll be able to set custom cookies for any specific scenario!


garris:
closes #442 
closes #443 
closes #421

PR #453, opened 2017-06-01
Allow absolute paths for snapshot directories

ktsn:
Hi, thank you for the great tool! I'm using it in some projects and it is really helpful!

I'm making this PR because I realized BackstopJS does not handle absolute paths in `paths.bitmaps_reference` and `paths.bitmaps_text`. This is not a problem in ordinary projects but I'm facing the case that needs to use absolute paths in it. This PR allows us to set absolute paths in the config.

You can confirm the current behavior by the following reproduction.
https://github.com/ktsn/backstop-compare-test

By the way, I'm not sure how we should write test code for such the fix. If it is needed, please let me know.

Thanks!

### DIFF
--- a/core/util/compare.js
+++ b/core/util/compare.js
@@ -69,8 +69,8 @@ module.exports = function (config) {
   return map(compareConfig.testPairs, function (pair) {
     var Test = report.addTest(pair);
 
-    var referencePath = path.join(config.projectPath, pair.reference);
-    var testPath = path.join(config.projectPath, pair.test);
+    var referencePath = path.resolve(config.projectPath, pair.reference);
+    var testPath = path.resolve(config.projectPath, pair.test);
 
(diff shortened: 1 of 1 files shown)


garris:
Thanks for the PR.  could you add this... https://github.com/ktsn/backstop-compare-test  to the examples directory.  Maybe name it absolute-paths.

ktsn:
@garris Added it into the `examples` directory with some tweaks.

garris:
Thank you for posting this.  I am planning to promote the current canary to @latest this week.  I will merge this into canary after promoting the current so we can test it out.   I am just a little concerned this could break existing implementations so I want others to test it for a while.  Cheers.

garris:
The files have changed significantly so I will merge this by hand.  I will close the ticket when complete.

PR #467, opened 2017-06-28
Reverse the test and reference image

aisensiy:
Reverse the test and reference image to make it align with the Reference / Test button in Comparison result popup.


garris:
closed by https://github.com/garris/BackstopJS/commit/d7e242abd77912abfcde87bdac5efa63c3520e69

PR #482, opened 2017-07-21
Chromy legacy fix

justin-parker:
Resolve issues with path resolution when running in "Legacy Mode"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,14 @@
-before_install:
-  - export CHROME_BIN=/usr/bin/chromium
-  - export DISPLAY=:99.0
-  - sh -e /etc/init.d/xvfb start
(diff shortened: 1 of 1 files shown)


garris:
Hi @justin-parker -- I am planning on eliminating legacy mode very soon.   I looked at this PR and I am pretty sure of two things...
- any use-case which currently relies on legacy mode can be very easily migrated
- without legacy mode -- the changes in this PR may not be necessary

Before I remove legacy mode I wanted to contact you to try to make sure it won't be disruptive to you.  Can you please comment on your use case -- I'd like to see if it would be possible for you to migrate.

justin-parker:
I was able to add BacktopJS as a node module in my app.  I no longer require legacy support.

garris:
Thanks for following up.  Cheers.

PR #503, opened 2017-08-22
Chromy: Prevent uncaught errors and make them available for the parent application

YStrauch:
Hi!

Chromy errors are not rethrown correctly. i.E. if I call chromy.wait with an invalid selector in a scenario, my console only prints "(node:86933) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TimeoutError: timeout", but I can neither catch the error (because it is not rethrown properly) nor find additional information (like scenario or label). Since a timeout occures > 30 seconds asynchronously and I have hundreds of tests running at once, I have no idea what went wrong and am pretty much lost.

This commit adds a nice error that bubbles up so that an outer application may react. If no outer catch exists scenario and viewport details are shown in the console so the user knows where to look for the mistake.

### DIFF
--- a/core/util/BackstopException.js
+++ b/core/util/BackstopException.js
@@ -1,0 +1,15 @@
+module.exports = class BackstopException {
+  constructor(msg, scenario, viewport, originalError) {
+    this.msg = msg;
+    this.scenario = scenario;
+    this.viewport = viewport;
+    this.originalError = originalError;
+  }
+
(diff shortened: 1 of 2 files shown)


garris:
This looks great!

YStrauch:
Yay!

garris:
Merged.

PR #525, opened 2017-08-29
Accept chromePath in config file

kiran-redhat:
chrome's executable location can be specified with 'chromePath' property in BackstopJS config file.


garris:
Can we make a change?  How about changing `hostFlags` to `hostConfig`, then directly extend the chromy options object.   We should also make `hostFlags` backward compatible and have it extend `chromeFlags` if specified separately.   I think this keeps the api footprint small and flexible as new requirements are added.   What do you think?

kiran-redhat:
Sounds good..I will work on that..

kiran-redhat:
is this what you mean @garris ?

kiran-redhat:
Hey @garris, I've tested, this accepts chromePath, hostFlags and newly added engineConfig(happy to change it to hostConfig). I am happy to add engineConfig for Casper as well, your thoughts ?

garris:
@kiran-redhat  I'm sorry.   I didn't see your update!   I just pushed a commit that might clobber some of your changes.

I have some feedback on your PR.   I will follow up as soon as I can.  Thanks so much.

garris:
@kiran-redhat  here is a more detailed acceptance criteria.  Sorry for not being clearer before.   Hope this helps.


Behavior 
----

The engineConfig property will accept an object and be used as the chromy options argument.  (Line 164)

If no engineConfig property is set then a default configuration will be used.  

If engineConfig is provided then all property values supplied will be appended, or override, any default chromy option values that would otherwise be used.   The one exception to this is the chromeFlags property.

The chromeFlags property, will always be overwritten by the internally generated `flags` variable (see line 159).  However, engineConfig.chromeFlags will override engineFlags is both are present.   engineFlags should still work for backward compatibility.   If found -- please add a console log warning users that engineFlags is deprecated. 

the debugWindow flag should behave the same way as chromeFlags.  It will not be depricated. 

Implementation 
----
Create an internal default property and use object.assign() to extend the property-- which will be used as an input to new Chromy().

garris:
And thank you adding the extra flags -- I have added the ability to directly set any properties in chromy with an `engineOptions` property -- so now all options are available.

Please see and comment on https://github.com/garris/BackstopJS/pull/569/files#diff-04c6e90faac2675aa89e2176d2eec7d8

thank you!

kiran-redhat:
@garris , I think this looks good, I will test and let you know.

kiran-redhat:
This change is been done and merged to master from https://github.com/garris/BackstopJS/pull/569, so closing this.

PR #529, opened 2017-08-31
Fix makeSafe issue `TypeError: undefined is not an object (evaluating 'str.replace')`

reneolivo:
The `makeSafe` function would throw a `TypeError: undefined is not an object (evaluating 'str.replace')` error when the str argument is undefined. This could happen in the following scenarios:

* missing the *label* of the scenario.
* missing the *label* of the viewport in the config file.
* using *name* instead of *label* for the viewport. This could be an honest mistake, because even the example files are confused:
screenshot 2017-08-31 07 22 58


reneolivo:
Hello @garris, I amended my PR to add the `|| ''` code. Is this ok now?

garris:
Yes, looks good.  Thank you!

adbutterfield:
Hello! Any idea when this will get released? I just hit this issue today too.

garris:
very soon -- will go out with next release.

adbutterfield:
Awesome! Thanks!!!

garris:
npm published @v3.0.32

PR #534, opened 2017-09-01
filter for approve

anton-kulagin:


garris:
@anton-kulagin I just took your diff and made changes by hand.   It is on master.  Would you mind testing it?  Please let me know if it is working as expected.   Thanks!

garris:
If it is working -- it would be great if you could update the readme as well -- thank you for the help!

garris:
Verified.

Issue #541, opened 2017-09-08
"window.exists is not a function" after loading another page in a script


laubsauger:
Hi @garris,

i think there is a problem when a scenario script causes a new page to be loaded.

I'm trying to fill out a form on a page and submit the form afterwards (this works).
The form actions brings me to a new page where the screenshot is supposed to get taken.

As soon as the new page is loaded and backstop tries to capture the screenshot it throws an exception:
````bash
COMMAND | Command `reference` ended with an error after [5.807s]
COMMAND | BackstopException: Account-Debitorwahl on 320-smallestPhone: Error: An error has occurred evaluating the script in the browser.TypeError: window.exists is not a function
    at window._backstopSelectorsExpMap.window._backstopSelectorsExp.reduce (<anonymous>:13:28)
    at Array.reduce (<anonymous>)
    at result (<anonymous>:11:72)
    at <anonymous>:22:9
````
I had a quick look around and realised that this function is provided by `backstopTools.js` which is injected after the initial goToUrl via Chromy. They are never re-injected when loading another page.

I quickly tried to manually inject `backstopTools.js` via chromy after the page load and that fixes the issue.

See my onReadyScript for this scenario below

````javascript
const ConfigProvider = require('../../../_scripts/configProvider.js');
const path = require('path');

module.exports = (chromy, scenario, vp) => {
    const config = new ConfigProvider(scenario.url);

    chromy.type('input#email', config.getUsername());
    chromy.type('input#password', config.getPassword());
    chromy.click('form[action="/action/account/login"] input[type="submit"]', { waitLoadEvent: true });

    // Inject BackstopTools to provide helper methods needed for screenshots
    const BACKSTOP_TOOLS_PATH = path.join(__dirname, '../../../node_modules/backstopjs') + '/capture/backstopTools.js';
    chromy.inject('js', BACKSTOP_TOOLS_PATH);

    return chromy;
};
````

Backstop 3.0.19
Node 8.4.0
Chrome 61.0.3163.79 (Official Build) (64-bit)
CentOS7

garris:
@laubsauger Thanks for posting this!   Yes.  This is something that I really want to fix.   I started reorganizing the code to make this happen -- but havent been able to finish it.

If you pull the new version you will see that file has been jammed into `runChromy.js`

https://github.com/garris/BackstopJS/blob/bc4dfc598bd3e9a40530713a845e820127bfdf5b/core/util/runChromy.js#L251-L258

[...]

PR #550, opened 2017-09-20
Inject backstopTools

laubsauger:
Solves #541.

I would be glad if somebody could enlighten me why this works. 

Using chromy's `inject` method resulted in no joy at all. Don't get why. Worked in other cases for me.
Trying to do this via `chromy.on('Page.loadEventFired')...` also didn't work.
I guess because the first evaluations are using these methods before this event?

### DIFF
--- a/core/util/runChromy.js
+++ b/core/util/runChromy.js
@@ -9,7 +9,6 @@ const CHROMY_STARTING_PORT_NUMBER = 9222;
 const DEFAULT_FILENAME_TEMPLATE = '{configId}_{scenarioLabel}_{selectorIndex}_{selectorLabel}_{viewportIndex}_{viewportLabel}';
 const DEFAULT_BITMAPS_TEST_DIR = 'bitmaps_test';
 const DEFAULT_BITMAPS_REFERENCE_DIR = 'bitmaps_reference';
-// const BACKSTOP_TOOLS_PATH = '/capture/backstopTools.js';
 const SELECTOR_NOT_FOUND_PATH = '/capture/resources/selectorNotFound_noun_164558_cc.png';
 const HIDDEN_SELECTOR_PATH = '/capture/resources/hiddenSelector_noun_63405.png';
 const BODY_SELECTOR = 'body';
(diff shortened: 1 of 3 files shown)


core/util/runChromy.js
```diff
@@ -9,114 +9,16 @@ const CHROMY_STARTING_PORT_NUMBER = 9222;
...
 const HIDDEN_SELECTOR_PATH = '/capture/resources/hiddenSelector_noun_63405.png';
 const BODY_SELECTOR = 'body';
 const DOCUMENT_SELECTOR = 'document';
 const NOCLIP_SELECTOR = 'body:noclip';
 const VIEWPORT_SELECTOR = 'viewport';
 
+const injectBackstopTools = require('../../capture/backstopTools.js');
 const BackstopException = require('../util/BackstopException.js');
 
-// ============================== MOVE TO NEW FILE VVVVV ===========================
```
garris: Please move these functions into the new file include verbatim -- these have been tested.  Let's not use the original backstopTools methods.  Thanks.

core/util/runChromy.js
```diff
@@ -347,6 +232,8 @@ function processScenarioView (scenario, variantOrScenarioLabelSafe, scenarioLabe
   }
 
   return new Promise((resolve, reject) => {
+    injectBackstopTools(chromy);
```
garris: injectBackstopTools() should probably be moved to line 211 -- and only run if `!window.exists`.   Does that work?

garris: injectBackstopTools() probably also unconditionally needs to run on line 151.

laubsauger:
You're right. This kills readyEvent/readySelector. Slight oversight on my part.

core/util/runChromy.js
laubsauger: I'm not sure i understand your request. You want those methods from `runChromy.js` to be moved to `backstopTools.js` as is? Isn't that the case? I saw no difference aside from the usage of constants in for the original methods in `backstopTools.js`.

garris: Thanks for looking at this.  Yes, That is the request.   I think there may have been tweaks to the methods inside of `runChromy.js`  -- That code has been working in prod -- so I'd like to ensure there are as few changes as possible.    I think there may be some minimal inline documentation added as well.   Simply copying the whole thing may be easier than attempting to diff.

garris: We may also want to extend the names to avoid any collisions.  I never liked the name .exists -- it's way too common.  maybe `window._backstopTools.exists()` would be better.

garris:
No worries!

laubsauger:
I've moved the injection to the suggested places, moved backstopTools method as is to a new property to avoid conflicts.
Your test config is working now.

core/util/runChromy.js
```diff
@@ -247,30 +149,15 @@ function processScenarioView (scenario, variantOrScenarioLabelSafe, scenarioLabe
...
-  chromy.defineFunction(hasLogged);
-  chromy.evaluate(_ => startConsoleLogger());
-  // console.log('Loading >>>' + config.env.backstop + BACKSTOP_TOOLS_PATH);
-  // chromy.inject('js', config.env.backstop + BACKSTOP_TOOLS_PATH);
-
-  // chromy.evaluate(() => console.log(exists))
-  // chromy.evaluate(() => console.log("window.exists ^^^^"))
-
-// =================UPDATE ^^^^^^^^^ ==================
+  injectBackstopTools(chromy);
```
laubsauger: I've tried to do it conditionally but it failed then.

````javascript
chromy.evaluate(() => {
    return !window._backstopTools;
}).result(_result => {
    if (_result) {
      injectBackstopTools(chromy)
    }
});
````
I don't have time left today to figure this out.

garris:
This is awesome.  Thank you!

Issue #555, opened 2017-10-05
Bug with getting scenarios list


anton-kulagin:
I have express js server which on get request start backstop test. Also i can send filter parameter.  See example above
```
router.get('/', function(req, res, next) {
    let filterVal = req.query.filter || "";
        backstopDef = backstop('test', {
            filter: filterVal
        }) .then(function(val) {
            res.json({ "answer": val})
        })
        .catch(function(reason) {
            res.json({ "answer": 'Error ' + reason;})
        })
});
```
And the problem is: When i make  request with filter it filtered list of scenarios correct , but when i make next request without filter i got wrong list of scenarios. 
Step to reproduce:
1. I have 2 test scenario
2. I make request and send filter parameter
3. I got correct result
4. I make request without filters
5. I got incorrect list of scenarious
I spend few hours yesterday to find why it happens and i`m sure that problem in createBitmaps.js ->decorateConfigForCapture
```
 if (typeof config.args.config === 'object') {
    configJSON = config.args.config;
  } else {
    configJSON = require(config.backstopConfigFileName);   //<- Here we got not an copy of data but reference
  }
```
And when we do it with filter 
```
config.args.filter.split(',').forEach(function (filteredTest) {
      each(configJSON.scenarios, function (scenario) {
        if (regexTest(scenario.label, filteredTest)) {
          scenarii.push(scenario);
        }
      });
    });
    configJSON.scenarios = scenarii;
```
we change reference value and on next call this function we got previously filtered scenarios.
I fix it just implemented copying content of config.backstopConfigFileName
from
```
configJSON = require(config.backstopConfigFileName);
```
To
```
 configJSON = Object.assign({}, require(config.backstopConfigFileName));
```
It helps me to fix my issue and shouldn`t  break anything else. It will be great if you can include this fix to library

anton-kulagin:
Hi @garris, could you look into it.

garris:
@anton-kulagin   Yes,  your thinking looks very good.  Please go ahead and submit a PR!

Also your project looks VERY COOL.   I love that you are integrating into express like that -- great idea!   It would be wonderful if you have a small example you would like to post to the `<backstopjs>/examples` directory -- I am sure other developers would benefit from seeing the example.  Cheers!

anton-kulagin:
@garris Hi, could you look into pull request for this issue?

PR #558, opened 2017-10-08
updated get config json #555 issue

anton-kulagin:
Fix according #555 issue


anton-kulagin:
@garris Hi,could you look into it when you be able. Without this update work with expressjs has an issue

garris:
This is merged and published.  @3.0.30

PR #565, opened 2017-10-13
add hyperlink to open diff in new tab

jparkerweb:


garris:
There you go.

garris:
Thanks!

PR #569, opened 2017-10-18
add engineOptions to config

garris:
Enables setting/overriding of Chromy options.

**Please note:**
- Setting `port` is _very_ not advised.
- Setting `chromeFlags` will override all chromeFlags properties set by backstop -- EXCEPT FOR `--window-size`...  (`--window-size` flag will be added by backstop if not found in chromeFlags) 
- Setting `--window-size` explicitly in `chromeFlags` will override values used in your viewport settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,6 +531,14 @@ An example config below...
 
 ### Integration options (local install)
 
+TLDR; run the example here...
+```
+cd backstopjs/test/configs/
+node multi_step node_example
+```
+
+Details...
+
 Installing BackstopJS locally to your project makes a few integration options available.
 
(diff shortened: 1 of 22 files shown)


VladimirPittner:
I have got same error as @kiran-redhat. 
Config:
```
...
engine: "chrome",
engineOptions: {
    waitTimeout: 10000,
},
...
```

garris:
Yes.  sorry guys -- I just got a chance to test.  Something is broken in there.   If you have time to take a look it would be a big help.  Otherwise I will get to it as soon as I can.

PR #579, opened 2017-10-25
Update makeConfig.js

anton-kulagin:
Remove backstop config from cache to provide ability to change it from server side(expressjs as example) and get updated version after that

### DIFF
--- a/core/util/makeConfig.js
+++ b/core/util/makeConfig.js
@@ -40,6 +40,8 @@ function loadProjectConfig (command, options, config) {
       userConfig = options.config;
     } else if (config.backstopConfigFileName) {
       try {
+        //Remove from cache config content
+ +      delete require.cache[require.resolve(config.backstopConfigFileName)]
         console.log('Loading config: ', config.backstopConfigFileName, '\n');
         userConfig = require(config.backstopConfigFileName);
       } catch (e) {


anton-kulagin:
Hi @garris . Could you look into it?

core/util/makeConfig.js
```diff
@@ -40,6 +40,8 @@ function loadProjectConfig (command, options, config) {
       userConfig = options.config;
     } else if (config.backstopConfigFileName) {
       try {
+        //Remove from cache config content
+ +      delete require.cache[require.resolve(config.backstopConfigFileName)]
```
garris: @anton-kulagin Sorry for the delay.  Very very busy at work.   I am not very familiar with this section of code and I am not sure what the side-effects will be.   Can you give me some context?

anton-kulagin: @garris Sure, so my idea is to provide possibility to write test scenarios from browser. I use angular + express. So i send to express server new scenario and it should update backstop.json file. And everything fine with this. But after this update backstop still see outdated list of test and this list will be updated just after restarting express server because of when you use `require` inside makeConfig.js it create cache for config file. 
**By adding this one line of code i will just say for require method to take backstop.json file each time by reading it from drive but not from cache**

garris:
Merged.

PR #605, opened 2017-11-22
Be more specific

trungdq88:
Actually showing the user config if detected in `options`.

### DIFF
--- a/core/util/makeConfig.js
+++ b/core/util/makeConfig.js
@@ -36,7 +36,7 @@ function loadProjectConfig (command, options, config) {
   var CMD_REQUIRES_CONFIG = command !== 'genConfig';
   if (CMD_REQUIRES_CONFIG) {
     if (options && typeof options.config === 'object') {
-      console.log('Config detected:');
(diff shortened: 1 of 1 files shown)


core/util/makeConfig.js
```diff
@@ -36,7 +36,7 @@ function loadProjectConfig (command, options, config) {
   var CMD_REQUIRES_CONFIG = command !== 'genConfig';
   if (CMD_REQUIRES_CONFIG) {
     if (options && typeof options.config === 'object') {
-      console.log('Config detected:');
+      console.log('Config detected:', options.config);
```
garris: hmmm.  I like this but what if this is a totally huge config?  That could get annoying -- maybe wrap this in an `if(config.debug){}` condition?  Also, if you're going to do something like this I would prefer it formatted via `JSON.stringify(config, null, 2)`

otherwise looks good.

trungdq88: That's reasonable. Will fix.

garris:
Merged!   Thanks for your patience!

PR #615, opened 2017-12-14
Upgrade chromy from 0.5.5 to 0.5.7

kiran-redhat:


tbal:
The package-lock.json should be updated too, no?

Issue #602, opened 2017-11-21
Tagging docker images


samuelhwilliams:
At the moment it looks like BackstopJS only records a single docker image tag: latest. This means it's not easy to use a consistent, reproducible image without copying the Dockerfile and using that to generate a static image.

It would be great if this project could start tagging releases with unique references, as well as updating latest, so that those who want to (e.g. me) can reference a specific build to minimise the risk of an update to this library inadvertently breaking our tests in some way.

garris:
I still don't know what I'm doing with docker. 😔   

what needs to happen?   Can you give me the steps?   It's probably pretty easy right?

samuelhwilliams:
What do you currently do to build and push a new docker image? I don't see the Dockerfile in this repo; do you have it somewhere else in source control?

I presume you run something like `docker build -t backstopjs/backstopjs .`. This will default to building an image tagged as `latest`, and when you push this up, it'll overwrite the previous image tagged as `latest`.

The simplest thing to do is start explicitly including tags when you build and push your docker images. You can do this like so: `docker build -t backstopjs/backstopjs:3.0.32 .`. This will create an image tagged as `3.0.32` (the BackstopJS version used in the latest Docker image, which is probably a reasonable place to start, unless you expect to do ever update the dockerfile without updating the Backstop version). You can then push it up with `docker push backstopjs/backstopjs:3.0.32`.

The `latest` tag doesn't necessarily do what you expect, insofar as it doesn't track the latest image you've pushed. It's just another tag name. So if you also want to update latest, you could instead run `docker build -t backstopjs/backstopjs:3.0.32 -t backstopjs/backstopjs:latest .` and then push up both tags (separately as you can't push multiple tags at once).

This would help people integrate a specific version of Backstop rather than having to track latest or copy your Dockerfile and build the image locally.

samuelhwilliams:
Apologies, just noticed this is already open in https://github.com/garris/BackstopJS/issues/562.

I will have a think about how to tie this into your hook already so that it can be handled automatically.

samuelhwilliams:
@garris Ok, so I assume you're using Docker Hub's automated builds, configured at https://hub.docker.com/r/backstopjs/backstopjs/~/settings/automated-builds.

[...]

PR #617, opened 2017-12-15
Tagging docker images - #602

Baltazardoung:
See https://github.com/garris/BackstopJS/issues/602#issuecomment-351777465
* Update Dockerfile and add the build hook to pass the version as
argument (extracted from the package.json file)
* Add the "post_push" hook (to generate the "latest" image)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,10 +1,12 @@
 FROM node:8.5.0-alpine
 
+ARG BACKSTOPJS_VERSION
+
 ENV \
 	PHANTOMJS_VERSION=2.1.1 \
 	CASPERJS_VERSION=1.1.4 \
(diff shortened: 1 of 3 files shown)


garris:
Beautiful— thank you.

samuelhwilliams:
One thing I've noticed from this, actually, is that it would probably be sensible to pass in the versions for the other dependencies (eg phantom and casper), as the ones specified in the Dockerfile seem to be out of sync with the `package-lock.json` file. PhantomJS is locked at 2.1.14 but this Dockerfile uses 2.1.1. If you're happy I'll go ahead and open another PR with a patch @garris?

garris:
Yes — that would be great!  Thank you!

Issue #608, opened 2017-12-01
How can i access Chromy.addCustomDevice?


deap82:
How can I access the static Chromy.addCustomDevice function?
I tried to  do

> const Chromy = require('chromy');

in my onReady.js, but just importing it like that causes the "backstop reference" command to fail.

Any leads appreciated.

garris:
Please take a look at the example scripts.  You don’t need to “require” Chromy.  It’s passed as an argument to the callback closure which your script runs in.

deap82:
@garris I did have a look at the example scripts but can't find an example for this. This specific method - addCustomDevice - is not available on the chromy instance, it's static on the type.

If I want to use

`chromy.emulate('myCustomDeviceConfig');`

I need to first call

`Chromy.addCustomDevice({ name: 'myCustomDeviceConfig', /.../ });`

Notice **c**hromy vs **C**hromy.

If I do
`console.log('chromy.emulate', chromy.emulate);`
I can see the body of the emulate function logged out to console.

If I do
`console.log('chromy.addCustomDevice', chromy.addCustomDevice);`
I get an exception.

garris:
Oh, didn’t notice that. If it’s a static method you can just import it.  But you might need to alias it to avoid collisions.  Does that work?

deap82:
I created a small gulp task and when I run this I get the function logged out:

```
gulp.task('backstop-foo', function (cb) {
    var ChromyStatic = require('chromy');
    console.log('ChromyStatic.addCustomDevice', ChromyStatic.addCustomDevice);
    cb();
});
```

But when I try this in my onReady.js for backstop:
```
module.exports = function (chromy, scenario, vp) {
    var ChromyStatic = require('chromy');
    console.log('ChromyStatic', ChromyStatic);
};
```
I get an error when I run "backstop reference", this is the output I get in console:

> BackstopJS v3.0.32
> Loading config:  C:\\[SOME-PATH]\backstop-settings.js
> COMMAND | Executing core for `reference`
> createBitmaps | Selected 1 of 1 scenarios.
> Starting Chromy: {"chromeFlags":["--disable-gpu","--force-device-scale-factor=1","--disable-infobars=true","--window-size=375,667"],"port":9222,"waitTimeout":30000,"visible":false}
> CREATING NEW REFERENCE FILES
>       COMMAND | Command `reference` ended with an error after [0.572s]

I'm running "backstop reference" with the --configPath flag, so I tried to access it from that .js file like this:
```
[...]

PR #621, opened 2017-12-22
Access to static Chromy functions

deap82:
- Access to static chromy reference in onBefore/onReady scripts. Example usage in README.md.
- Core: Always log out error messages.

Related to: https://github.com/garris/BackstopJS/issues/608


garris:
Merged! Thanks for your patience!

PR #629, opened 2018-01-16
Add option to use resemble`s ignoreAntialiasing()

zbycz:
Hi, 

I would like an option to ignore antialiasing. In resemble it is possible by 
`resemble().compareTo().ignoreAntialiasing()`. So i've added the configuration to the `resembleOutputOptions` in backstop.json.

Also I updated the react-example to feature this option. Would you be ok with merging and possibly releasing?

Thanks for great tool!


zbycz:
Cheers! Any idea on when could be the next release? :-)

garris:
How about right now -- does that work?  😉

Just published backstopjs@3.0.38

Thank you for your PR!

zbycz:
So cool 👍

PR #645, opened 2018-01-28
Add PowerShell cmd to kill headless chrome instances

liamaharon:
Just one way we could integrate the PowerShell command. 

Feel free to change if you'd prefer different formatting or wording.


garris:
Great — thank you!

PR #646, opened 2018-01-28
Fixs

mantovanig:
- hide Ref/test/scrubber buttons if there is no diff
- give the Ref/test/scrubber buttons a selected state
- responsive image scrubber


garris:
Awesome!

Issue #603, opened 2017-11-21
BackstopJS hangs in Docker container


questionmarc:
I seem to keep having this problem where the chrome headless hangs forever until the script ends. Here's the output I'm seeing:

```bash
Starting Chromy: {"chromeFlags":["--disable-gpu","--force-device-scale-factor=1","--disable-infobars=true","--no-sandbox","--window-size=320,480"],"port":9222,"waitTimeout":30000,"visible":false}
Starting Chromy: {"chromeFlags":["--disable-gpu","--force-device-scale-factor=1","--disable-infobars=true","--no-sandbox","--window-size=1024,768"],"port":9223,"waitTimeout":30000,"visible":false}
Starting Chromy: {"chromeFlags":["--disable-gpu","--force-device-scale-factor=1","--disable-infobars=true","--no-sandbox","--window-size=1600,900"],"port":9224,"waitTimeout":30000,"visible":false}
9223 Chrome v61 detected.
9223 ***WARNING! CHROME VERSION 62 OR GREATER IS REQUIRED. PLEASE UPDATE YOUR CHROME APP!***
9222 Chrome v61 detected.
9222 ***WARNING! CHROME VERSION 62 OR GREATER IS REQUIRED. PLEASE UPDATE YOUR CHROME APP!***
9224 Chrome v61 detected.
9224 ***WARNING! CHROME VERSION 62 OR GREATER IS REQUIRED. PLEASE UPDATE YOUR CHROME APP!***
bash-4.3#
```

This is utilizing the same dockerfile as the one listed in the repo. I'm not sure why it's hanging but I don't get any real output. Changing the engine to PhantomJS works, and from looking at the debug output from the `debug` flag, it is reaching the webpage. I tried adding some flags to the `docker run` from reading this project: https://github.com/yukinying/chrome-headless-browser-docker . I also tried to add the `--no-sandbox` option to the `chromeFlags` but to no avail.

So I'm not sure if this is a chrome-headless issue? An issue with chromium v61? Alpine? I'm also not sure how to debug this further.

If you have any insight as to things I could be doing to make progress, please let me know.

questionmarc:
New Development: with debug on and setting the `asyncCaptureLimit` to `1`, I got one of the scenarios to complete. But then the other two time out.

questionmarc:
output for that run was something like
```bash
[...]

Issue #537, opened 2017-09-02
asyncCaptureLimit and asyncCompareLimit seems not working


ryuran:
After an error on Travis (https://github.com/OnetapInc/chromy/issues/62), I try to reduce async limit to pass test.
```
asyncCaptureLimit: 5,
asyncCompareLimit: 50
```

But nothing change 😢 : https://travis-ci.org/20minutes/colette/builds/271137302

garris:
Did you try to reduce to 1?

ryuran:
@garris yes my second try is at 1:1

garris:
I have just verified that asyncCaptureLimit and asyncCompareLimit are working as designed.  At this point I think you need to look at why your particular test is failing.

ryuran:
I add log for test scenario label.
It seems fail on `./docs/item-theme-skins.html` or the next one `./docs/item-theme-styles.html`.
I check this files with html validator, they are valid.
So I’m blocked. All is fine in local environment.

kiran-redhat:
@ryuran 1. are you about to find that in chrome console ? 

2. Do you see sale error with asyncCaptureLimit: 1,?

If 1 &2 are yes, clone  latest master and test. asyncCaptureLimit and asyncCompareLimit we're working fine for me

ryuran:
@kiran-redhat 
1. I don’t know how to return more info from browser from travis-CI
2. This travis build is running with `asyncCaptureLimit: 1` https://travis-ci.org/20minutes/colette/builds/271948792 I have the error

kiran-redhat:
for number 1: open your app in Chrome browser press f12 then click on
console then copy and paste your selector and hit enter. if you find element

wait for your selector to be visible (probably chromy.visible() )

ryuran:
No error in my local chrome.
Backstop.js run fine in local environment with all test to green.

I update backstopjs:
```
└─┬ backstopjs@3.0.24
  └── chromy@0.5.2
```
And I try to change delay to `200` (it was `100`), but it stops always on the same test.
Without error this time but with travis timeout:
https://travis-ci.org/20minutes/colette/builds/271982035

ryuran:
It seems fail most often on test with `<img>`

reynaldoapb:
Try to kill chrome with shell command  `taskkill /F /IM chrome.exe`  and run everything again.

axe312ger:
Turning it down to 
```
asyncCaptureLimit: 1,
asyncCompareLimit: 1
[...]

PR #668, opened 2018-02-13
Fixing Docker build issues. Updated chrome, Potentially fixing issues #603 #537.

iain17:
I wanted to use Chrome and the latest version of BackstopJS using Docker. However the automatic build of BackstopJS isn't working.

This PR fixes the docker build error and updates PhantomJS and Chrome. In order to run on the latest stable Chrome 64 build, I had to change the original alpine base image to the default node image. This potentially fixies the memory leak issues reported in #603 #537 

You can test it out on: https://hub.docker.com/r/iain17/backstopjs/

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,9 @@
-FROM node:8.5.0-alpine
+FROM node:8.5.0
 
 ARG BACKSTOPJS_VERSION
 
 ENV \
-	PHANTOMJS_VERSION=2.1.1 \
+	PHANTOMJS_VERSION=2.1.7 \
 	CASPERJS_VERSION=1.1.4 \
 	SLIMERJS_VERSION=0.10.3 \
 	BACKSTOPJS_VERSION=$BACKSTOPJS_VERSION \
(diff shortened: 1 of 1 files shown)


lmakarov:
The switch to the default (non-alpine) flavor of the base nodejs docker image inflated the resulting (compressed) image size from 303 MB to 653 MB - http://take.ms/9XvDE

Looks like Alpine Edge now has Chromium v64 available:
https://pkgs.alpinelinux.org/package/edge/community/x86_64/chromium

For a smaller image size either `alpine` or the `slim` flavor of the node image should be used.
See https://hub.docker.com/r/library/node/tags/

- `8` - Compressed size: 270 MB
- `8-slim` - Compressed size: 92 MB
- `8-alpine` - Compressed size: 23 MB

garris:
Hi @lmakarov!   Would you mind PR-ing?

lmakarov:
Hey @garris! I believe rolling back to Alpine would be the easiest. I do, however, like how much simpler the Dockerfile is with Debian. Alpine is the goto distribution for Docker images, however it does make things more complex, as weird workarounds have to be used quite often.

Using `8-slim` (a stripped down version of Debian) may be a good middle ground so long as it actually works without introducing workarounds of its own.

@iain17 would you be interested in trying `8-slim` as the base image? If not, I'll try to find some time this week for that.

PR #693, opened 2018-03-02
Feature/show url on text details component

matthiaskomarek:
* added url and reference url to TextDetails Component
* added url and reference url to Chromy test pairs

bildschirmfoto 2018-03-02 um 13 32 59

### DIFF
--- a/compare/output/index_bundle.js
+++ b/compare/output/index_bundle.js
@@ -13242,6 +13242,10 @@ var _DiffDetails = __webpack_require__(272);
 
 var _DiffDetails2 = _interopRequireDefault(_DiffDetails);
 
+var _UrlDetails = __webpack_require__(273);
+
+var _UrlDetails2 = _interopRequireDefault(_UrlDetails);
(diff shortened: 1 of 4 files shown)


garris:
@matthiaskomarek Thank you for your patience and thank you for your PR.

Please take a look at this mock. I would like you to add the `test url` (and the optional `ref url`) laid out in the following way.   Below is the current design.  Above is the proposed change.  Please shorten the links so that the link text can flow onto the same line as the optional DIFF results.  They should be styled such that the `url` heading should be grey (like the other headings) and the two links can inherit your existing link style...
image

matthiaskomarek:
Hey @garris thanks for your feedback. I have updated the PR accordingly.
I have also created a new Component for the `UrlDetails` and had to do some minor changes to the `DiffDetails` markup so that both components can be in the same `Row`

Please let me know what you think! Thanks

bildschirmfoto 2018-03-16 um 09 57 38

PR #718, opened 2018-03-27
Improve Puppeteer error handling

krisdigital:
* Add error images for Puppeteer selector capturing
* improve handling Puppeteer errors

### DIFF
--- a/core/util/runPuppet.js
+++ b/core/util/runPuppet.js
@@ -7,6 +7,8 @@ const ensureDirectoryPath = require('./ensureDirectoryPath');
 const injectBackstopTools = require('../../capture/backstopTools.js');
 const engineTools = require('./engineTools');
 
+const BackstopException = require('../util/BackstopException.js');
+
 const MIN_CHROME_VERSION = 62;
 const TEST_TIMEOUT = 60000;
 const DEFAULT_FILENAME_TEMPLATE = '{configId}_{scenarioLabel}_{selectorIndex}_{selectorLabel}_{viewportIndex}_{viewportLabel}';
@@ -29,7 +31,7 @@ module.exports = function (args) {
   const scenarioLabelSafe = engineTools.makeSafe(scenario.label);
   const variantOrScenarioLabelSafe = scenario._parent ? engineTools.makeSafe(scenario._parent.label) : scenarioLabelSafe;
 
-  return processScenarioView(scenario, variantOrScenarioLabelSafe, scenarioLabelSafe, viewport, config, runId, assignedPort);
+  return processScenarioView(scenario, variantOrScenarioLabelSafe, scenarioLabelSafe, viewport, config, runId);
 };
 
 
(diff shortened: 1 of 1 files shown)


core/util/runPuppet.js
```diff
@@ -199,9 +204,20 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
         backstopSelectorsExpMap: window._backstopSelectorsExpMap
       };
     })
-
-  return new Promise((resolve, reject) => {
-    resolve(delegateSelectors(
+  }
+  
+  await puppetCommands().catch(e => {
+    result = {
```
garris: Should this be a const?  It looks like this `result` is not the same `result` used above?  Or did I maybe read this wrong?

core/util/runPuppet.js
```diff
@@ -71,114 +74,116 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
...
-  await injectBackstopTools(page);
-
-  //  --- WAIT FOR READY EVENT ---
-  var readyEvent = scenario.readyEvent || config.readyEvent;
-  if (readyEvent) {
-    await page.evaluate(`window._readyEvent = '${readyEvent}'`)
+    //  --- OPEN URL ---
+    var url = translateUrl(scenario.url);
+    if (isReference && scenario.referenceUrl) {
+      url = scenario.referenceUrl;
```
garris: Holy crap -- I just found a bug that I introduced a few days ago!   This value should also be wrapped in a `translateUrl `   -- would you mind adding that?

`url = translateUrl(scenario.referenceUrl);`

core/util/runPuppet.js
krisdigital: it is actually the same result.. If the puppeteer commands fail the result is empty and this was the best I could come up with without refactoring too much.. 🛠

core/util/runPuppet.js
krisdigital: done!

core/util/runPuppet.js
```diff
@@ -71,114 +74,116 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
...
-    var beforeScriptPath = path.resolve(engineScriptsPath, onBeforeScript);
-    if (fs.existsSync(beforeScriptPath)) {
-      await require(beforeScriptPath)(page, scenario, viewport, isReference, browser);
-    } else {
-      console.warn('WARNING: script not found: ' + beforeScriptPath);
+  
+  let result;
+  const puppetCommands = async () => {
+    // --- BEFORE SCRIPT ---
+    var onBeforeScript = scenario.onBeforeScript || config.onBeforeScript;
```
matthiaskomarek: Is there a reason for using `var`?

core/util/runPuppet.js
```diff
@@ -71,114 +74,116 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
...
+    if (scenario.hasOwnProperty('removeSelectors')) {
+      const removeSelectors = async () => {
+        return Promise.all(
+          scenario.removeSelectors.map(async (selector) => {
+            await page
+              .evaluate(`window._backstopSelector = '${selector}'`);
+
+            await page
+              .evaluate(() => {
+                Array.prototype.forEach.call(document.querySelectorAll(window._backstopSelector), function (s, j) {
```
matthiaskomarek: This could also be `Array.from(document.querySelectorAll(...).forEach()` its easier to understand.
`j` is also unused

core/util/runPuppet.js
```diff
@@ -211,12 +227,22 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
...
+    )
+  } catch (e) {
+    error = e;
+  }
+  
+  return new Promise((resolve, reject) => {
+    if (compareConfig) {
+      resolve(compareConfig);
+    } else {
+      resolve(new BackstopException('Puppeteer error', scenario, viewport, error));
```
matthiaskomarek: Shouldn't this be `reject` instead of `resolve`? Because now, the promise resolves with an error, which should be a reject or?

core/util/runPuppet.js
krisdigital: the reason is that I copy pasted it from the chromy script 😜

core/util/runPuppet.js
garris: In this case we resolve either way.   The intention is to pass the error up as a result.

core/util/runPuppet.js
garris: Cool.  Thanks!

garris:
Great work!  Big big thanks! 🙏

PR #719, opened 2018-03-30
README: Troubleshooting-- edit example command. Also some typo fixes.

aliceyoung9:
I thought it'd be good to change that example install command in troubleshooting, because a new dev might copy-and-paste that exact line and install the wrong package entirely. Also did some spellchecking.

Thanks!


garris:
Thank you ever so! 😺

PR #722, opened 2018-04-05
Adds option to pass additional args to puppeteer.

VesterDe:
This refers mostly to
https://github.com/GoogleChrome/puppeteer/issues/2161 which I can not
solve unless I pass --no-sandbox and --disable-setuid-sandbox to
puppeteer.

The correct syntax for the backstop.json config would then be (specifically for my use case, as an example):

`"engine": "puppet",`
`"puppeteerArgs": {args: ['--no-sandbox', '--disable-setuid-sandbox']},`

### DIFF
--- a/core/util/runPuppet.js
+++ b/core/util/runPuppet.js
@@ -56,7 +56,7 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
       ignoreHTTPSErrors: true,
       headless: !!!config.debugWindow
     },
-    config.puppeteerArgs
+    config.engineOptions
   );
 
   const browser = await puppeteer.launch(puppeteerArgs);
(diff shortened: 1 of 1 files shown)


cactusa:
@garris would this be considered as part of the puppeteer integration that was recently done?

garris:
@cactusa Yes, this is an important feature that should be added.

VesterDe:
Did I understand this correctly?

VesterDe:
Yeah, I was already using my branch to work with puppet, now I can change to the official repo again :)

Do you mean docs in readme.md? I can write a few sentences about these flags, but not the whole puppet guide, which I suppose would still need to be written.

garris:
Oh — you don’t have to write the entire puppet documentation.  Maybe just a note related to your PR that `engineOptions` can also be used to send a custom config object to puppeteer.  A short config object example would be helpful too.  Thanks!

garris:
... and yes,  in the readme — just to let others know that the feature is there.

PR #727, opened 2018-04-11
add new selector: activeSelector for chromy

silentHoo:
I needed an activeSelector so I want to share it with the community.


silentHoo:
I've added the implementation for casper and puppet but it's not yet tested because lack of time atm. Could anyone test the code? Best regards.

garris:
Let’s also add this to the backstop_feature test.

PR #747, opened 2018-04-19
Smoke test that runs in travis

KenCoder:
Don't merge yet - I intentionally put a console.error statement to see if this fails travis on the PR build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "build-compare": "webpack --config ./compare/webpack.config.js",
     "dev-compare": "webpack-dev-server --content-base ./compare/output --config ./compare/webpack.config.js",
     "integration-test": "rm -rf newdir && mkdir newdir && cd newdir && node ../cli/index.js genConfig && node ../cli/index.js reference && node ../cli/index.js test && node -e \"require('../')('test')\"",
-    "smoke-test": "cd test/configs/; node ../../cli/index.js test --config=backstop_features",
+    "smoke-test": "node test/smoke/smoke_test.js",
     "sanity-test": "cd test/configs/; node ../../cli/index.js test",
     "kill-zombies": "pkill -f \"(chrome)?(--headless)\"",
     "copy-report-bundle": "cp compare/output/index_bundle.js test/configs/backstop_data/html_report/",
--- a/test/configs/backstop_features.js
+++ b/test/configs/backstop_features.js
@@ -130,7 +130,7 @@ module.exports = {
     html_report: 'backstop_data/html_report',
     ci_report: 'backstop_data/ci_report'
   },
-  report: ['browser'],
+  report: ['CI'],
   engine: ENGINE,
   engineOptions: {},
   asyncCaptureLimit: 10,
--- a/test/smoke/smoke_test.js
+++ b/test/smoke/smoke_test.js
@@ -1,0 +1,70 @@
+const path = require('path');
+const childProcess = require('child_process');
+const util = require('util');
+
+// Catch errors from failing promises
+process.on('unhandledRejection', function (error) {
+  console.error(error && error.stack);
+  process.exit(3);
+});
+
+const child = childProcess.spawn(
+  path.join(__dirname, '../../cli/index.js'),
+  ['test', '--config=backstop_features'],
+  {cwd: path.join(__dirname, '../configs')}
+);
+
+const requiredStdout = [
+  'Element not found for capturing: .monkey',
+  'report | 44 Passed',
+  'report | 14 Failed'
+];
+
+const allowedStdErr = [
+  'Possible EventEmitter memory leak detected.'
+];
+
+const missingStdOut = {};
+requiredStdout.forEach(function (el) { missingStdOut[el] = false; });
+
+child.stdout.on('data', function (buf) {
+  const line = String(buf);
+  util.print(line);
+  const key = line.trim();
+  if (missingStdOut[key] === false) delete missingStdOut[key];
+});
+
+const stdErrs = [];
+child.stderr.on('data', function (buf) {
+  const line = String(buf);
+  util.print('StdErr:', line);
+  if (!allowedStdErr.filter(allowed => line.includes(allowed))) {
+    stdErrs.push(line.trim());
+  }
+});
+
+child.on('close', function (code) {
+  const problems = [];
+  const missing = requiredStdout.filter(function (el) { return missingStdOut[el]; });
+  if (missing.length !== 0) {
+    problems.push('The following line were expected in smoke test std output, but were not found:');
+    problems.push(...missing);
+  }
+
+  if (stdErrs.length !== 0) {
+    problems.push('Unexpected output on Std Error of smoke test:');
+    problems.push(...stdErrs);
+  }
+
+  if (code !== 1) {
+    problems.push(`Expected exit code 1, but it was ${code}`);
+  }
+  if (problems.length > 0) {
+    console.error('Smoke test FAILED:');
+    problems.forEach(line => console.error(line));
+    process.exit(1);
+  } else {
+    console.error('Smoke test succeeded.');
+    process.exit(0);
+  }
+});
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,4 @@ install:
   - npm install
 script:
   - npm run unit-test
+  - npm run smoke-test
--- a/core/util/runPuppet.js
+++ b/core/util/runPuppet.js
@@ -115,6 +115,8 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
       }
     }
 
+    console.error('This error is to test the travis build failing!');
+
     //  --- OPEN URL ---
     var url = scenario.url;
     if (isReference && scenario.referenceUrl) {


package.json
```diff
@@ -21,7 +21,7 @@
     "build-compare": "webpack --config ./compare/webpack.config.js",
     "dev-compare": "webpack-dev-server --content-base ./compare/output --config ./compare/webpack.config.js",
     "integration-test": "rm -rf newdir && mkdir newdir && cd newdir && node ../cli/index.js genConfig && node ../cli/index.js reference && node ../cli/index.js test && node -e \"require('../')('test')\"",
-    "smoke-test": "cd test/configs/; node ../../cli/index.js test --config=backstop_features",
+    "smoke-test": "node test/smoke/smoke_test.js",
```
garris: Very nice harness you built there!  

A suggestion: would you consider leaving the `smoke-test` command the same (so it executes the backstop_features test like before) -- and add a *new* command for your new harness.  Maybe call it like, `travis-test` or `git-commit-test` or `deploy-test`? 

I am totally open to suggestions -- my instinct is to keep it convenient for devs to still be able to run the raw smoke-test and see results (passing or non-passing). 

Does this make sense?

KenCoder: I think the harness works well for devs too because it tells them if it succeeds at the end. Right now it's hard to notice if something breaks (e.g., incorrect # of comparison failures or that a timeout). There are some scary looking messages in the output that I don't know if they mean somethings wrong or not (selector not found, etc).

The only difference is that the CI test doesn't pop up the browser report at the end. 

But I will add an argument so if you do
npm run smoke-test 

It will behave just like today with additional niceness and popup the browser, and 

npm run smoke-test -- ci

Will do it without the browser report.

Question for you: Will this have to run in Docker? It seems like launching in travis doesn't really work. If so I can't currently see the docker cloud config - maybe you can share it with me?

garris: Sounds good!

I honestly am not an expert in docker -- I just know enough to get me into trouble.  I am happy to share the configs with you -- they should all be in the `/docker` directory.

Also -- just so I am understanding correctly -- when you say...
> It seems like launching in travis doesn't really work
...are you saying you cant run backstop tests in travis?

KenCoder: > It seems like launching in travis doesn't really work

Well that's the way it is now - you can click below and see the error. I wasn't sure if you already knew this. If not - no worries I think I can fix this problem. So expect an updated PR shortly.

KenCoder:
Don't merge yet! I'm still working on it - unfortunately chrome in Linux produces image mismatch with chrome on the mac, so likely the mac version will have to use a docker container to generate the reference bitmaps. I will let you know when it's ready, and no worries about work - I know how it is.

garris:
😄 ok thank you.

BTW:  if you are running into rendering differences you could try adding this to the settings...
```
    "args": [
      "--no-sandbox",
      "--disable-setuid-sandbox",
      "--disable-gpu", 
      "--force-device-scale-factor=1", 
      "--disable-infobars=true"
    ]
```
In particular, —disable-gpu when used on Mac _and_ Linux may fix the font rendering issues.   It does seem to work across macOS boxes and _may_ work across Linux and Mac.

It would be good to know if this in fact can solve the issue.  I am happy to rerender using that setting if you don’t have a Mac handy.

And again, thanks so much for your effort 😊.

KenCoder:
Hey - any luck reviewing this? Any concerns?

garris:
Oh crap — sorry man.   So busy...   😿

I have a day to work on the library on 5/18 — but hopefully I can this  before that.

Thanks for your patience!

test/smoke/docker_smoke.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
garris: I would love to know what each of these lines did.  I find it hard to parse docker configs.   Particularly lines 5 & 6.   I guess I expect we would cd to a workspace, install backstop locally to that space and use the feature tests runner.   I don't understand why `/test/configs/linux/node_modules` is called out in there.

test/smoke/smoke_test.js
```diff
@@ -0,0 +1,102 @@
...
+function sleep (ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+async function runSmokeTest (description, commands) {
+  const child = childProcess.spawn(commands[0], commands.slice(1),
+    { cwd: configsDir, env: processEnv }
+  );
+
+  const requiredStdout = [
```
garris: This sort of thing should be placed at the top of this script with explicit comments around how to use this array to handle different test scenarios.

In this case we should name it something very clear.

test/smoke/smoke_test.js
```diff
@@ -0,0 +1,102 @@
...
+    'Element not found for capturing: .monkey',
+    'report | 0 Failed'
+  ];
+
+  const allowedStdErr = [
```
garris: same comment as above for this array.

test/smoke/smoke_test.js
```diff
@@ -0,0 +1,102 @@
...
+    'Possible EventEmitter memory leak detected.'
+  ];
+
+  let missingStdOut = [...requiredStdout];
+
+  child.stdout.on('data', function (buf) {
```
garris: Please use named functions for clarity.

test/smoke/smoke_test.js
```diff
@@ -0,0 +1,102 @@
...
+    const line = String(buf);
+    util.print('StdErr:', line);
+    if (allowedStdErr.filter(allowed => line.includes(allowed)).length === 0) {
+      stdErrs.push(line.trim());
+    }
+  });
+
+  const code = await new Promise(resolve => child.on('close', resolve));
+
+  const problems = [];
```
garris: would prefer if these were encapsulated a bit by functions so that the overall flow of this script can be easily grasped in a few lines.

assets/styles.css
```diff
@@ -43,7 +43,6 @@ a:hover {
   transform: translateZ(200px);
 }
 .lemurInYourFace {
-  transition: transform 750ms ease, opacity 750ms ease, visibility 800ms ease;
```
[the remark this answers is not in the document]
KenCoder: The way to do this is with the transitionend event. This could be a new backstop feature, or one can write custom Javascript to detect the transition and use the readyEvent feature of Backstop. There is already a test of readyEvent - this 'lemurInYourFace' is supposed to be the 'Simple' test case. Demonstrating CSS  transition in the "Simple" case seems wrong. 

How about we add another issue "demo CSS transition handling" that can be done in a separate PR? I didn't introduce this problem - I'm just cleaning up something that was already flaky.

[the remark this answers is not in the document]
garris: yes that works.  We can just comment the flaky test now and create an issue.

_There is actually an article on dwb  which discusses exactly your idea — using the transition event.  The onReady test script can attach a listener and just emit a console log which triggers the shot._

test/configs/backstop_features.js
```diff
@@ -1,5 +1,8 @@
+const process = require('process');
+
 const ENGINE = 'puppet';
 const SCRIPT_PATH = 'puppet';
+const PLATFORM = process.env.BACKSTOP_SMOKE_PLATFORM || (process.platform === 'darwin' ? 'osx' : 'linux');
 
 module.exports = {
   id: `${ENGINE}_backstop_features`,
```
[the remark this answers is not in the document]
KenCoder: Yes this is much nicer - I changed it to this.

test/configs/backstop_features.js
```diff
@@ -83,7 +86,7 @@ module.exports = {
     },
     {
       label: 'cookies',
-      cookiePath: 'backstop_data/cookies.json',
+      cookiePath: `${PLATFORM}/backstop_data/cookies.json`,
```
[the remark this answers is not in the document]
KenCoder: Fixed

test/configs/backstop_features.js
```diff
@@ -123,17 +126,18 @@ module.exports = {
     }
   ],
   paths: {
-    bitmaps_reference: 'backstop_data/bitmaps_reference',
-    bitmaps_test: 'backstop_data/bitmaps_test',
-    engine_scripts: 'backstop_data/engine_scripts',
-    html_report: 'backstop_data/html_report',
-    ci_report: 'backstop_data/ci_report'
+    bitmaps_reference: `${PLATFORM}/backstop_data/bitmaps_reference`,
```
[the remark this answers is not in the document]
KenCoder: Done

test/configs/backstop_features.js
```diff
@@ -125,15 +128,16 @@ module.exports = {
   paths: {
     bitmaps_reference: 'backstop_data/bitmaps_reference',
     bitmaps_test: 'backstop_data/bitmaps_test',
-    engine_scripts: 'backstop_data/engine_scripts',
+    engine_scripts: 'engine_scripts',
```
garris: Should also be under backstop data.

gabegorelick:
Is there anything preventing this from being merged, aside from needing a rebase? More tests would give contributors a lot of confidence their changes won't break things.

KenCoder:
We need to address #842 for the smoke tests to pass consistently. Then it needs to rebase.

PR #748, opened 2018-04-23
Adding peerDependencies for react

longstaff:
Moves react to peer dependency to prevent hard link and multiple react version installs.


garris:
Thanks for this!

PR #768, opened 2018-05-11
Adding peer dependencies and dev dependencies for react and react-dom

longstaff:
Moves react to peer dependency to prevent hard link and multiple react version installs.
Local development supported by adding dev dependency on react and react-dom.


garris:
Thanks!

PR #777, opened 2018-05-17
Use opn instead of open

ataylorme:
The [open package](https://www.npmjs.com/package/open) has not been updated in 4 years and has a [ciritical security vulnerability](https://nodesecurity.io/advisories/663). This replaces it with the [opn package](https://www.npmjs.com/package/opn) which is well maintained by the great [sindresorhus](https://www.npmjs.com/~sindresorhus) and does not have vulnerabilities.

PR #778, opened 2018-05-21
Ditch open

ataylorme:
Continuation of #777

@garris I figured it out. `opn` has a wait option that makes it so the promise doesn't resolve until the browser is closed. The wait option is `true` by default. After setting it to `false` the promise resolves immediately and things work as expected.

Since `opn` itself returns a promise I also removed the existing promise wrapped around it to avoid redundancy.


garris:
Ah ha!  👍👍   Two thumbs up.   Will pull this in as soon as I can.


Also — would love to hear about your use cases — or see presentations.  I’ve been wanting to get more of his stuff out there.

PR #790, opened 2018-06-03
[Docs] Jenkins Guide

Kenith:
Add Jenkins Guide into docker/Jenkins


garris:
Hi @kenith,  this looks good — but maybe this belongs in the examples directory — instead of under `./docker/`?

Kenith:
Hello @garris, we suggest to put it in `./docker/`, since the jenkins guide is based on the docker image `backstopjs/backstopjs:latest`. 

Also, I have added .dockerignore to make sure the Jenkins folder won't be added to docker image.

You are very welcome if you have any questions.

Thank you,
Ken

garris:
I am not sure that users will think the same way.  A _Jenkins_ user may actually expect to see the Docker config inside a top-level Jenkins directory (e.g. `./jenkins/docker/` or even `./jenkins-docker/`).  

But maybe more important, this Jenkins/DooD approach is a (very useful but) very opinionated implementation -- which seems to be very good as an example but perhaps not ready for production use because of the security concerns (e.g. giving Jenkins sudo permissions and docker-outside-of-docker) -- whereas the existing Docker implementation is intended for general use and therefore given 1st class status inside the repo.

I think this implementation serves as an **excellent** example -- it could probably be a very useful blog post as well.   I would gladly highlight this in the documentation -- I think it would be very very appreciated by Jenkins users.

Kenith:
Hello @garris,

Really apologies for delay. These days are lots of drama things.

I have moved the jenkins guide to examples.
And I fully agree with your opinions about the security issue. In the coming days, I will seek help from our dev ops team for more professional jenkins settings.

Cheers,
Ken

PR #812, opened 2018-06-22
fix fallback images in puppeteer (#810)

### LINKED ISSUE #810: [runPuppet] selectorShot uses wrong variable name
I think the ocurrences of `filePath` should be `path` -- filePath will be stuck on the last selector due to asynchronicity.

https://github.com/garris/BackstopJS/blob/742a2a0709ef3f41542bf148499fe812287b4834/core/util/runPuppet.js#L360-L376

This prevents creating the fallback images for failures, leading to false positives when both reference and test lack a selector.

pkra:
selectorShot: pass correct variable to fs.copy
Fixes #810

### DIFF
--- a/core/util/runPuppet.js
+++ b/core/util/runPuppet.js
@@ -368,11 +368,11 @@ async function captureScreenshot (page, browser, selector, selectorMap, config, 
           });
         } else {
           console.log(chalk.yellow(`Element not visible for capturing: ${s}`));
-          return fs.copy(config.env.backstop + HIDDEN_SELECTOR_PATH, filePath);
+          return fs.copy(config.env.backstop + HIDDEN_SELECTOR_PATH, path);
         }
       } else {
         console.log(chalk.magenta(`Element not found for capturing: ${s}`));
-        return fs.copy(config.env.backstop + SELECTOR_NOT_FOUND_PATH, filePath);
+        return fs.copy(config.env.backstop + SELECTOR_NOT_FOUND_PATH, path);
       }
     };
 


garris:
Sorry, I just reviewed this.  I believe this would break existing desired behavior.   Not sure what this fixes.   Thanks.

pkra:
>  I believe this would break existing desired behavior> 

Hm, that's odd. My impression was that `selectorShot` gets passed a `path` during `selectorsShot` (which uses `filePath` as variable name). If the element exists, `selectorShot` uses that `path` variable, but when it fails, it uses `filePath` -- which given the promises will be the last value of `filePath`. 

> Not sure what this fixes. 

I was trying to fix #810. I regularly have the same configuration applied to several URLs, some of which will not have particular elements. The current code will fail to generate fallback images in that case because generally `filePath !== path`.

pkra:
Minor edits and to add: the current code leads to false negatives in test results since no images are found (instead of both ref and test having the fallback image which are identical).

pkra:
@garris I've added a minimal example to #810. PTAL?

garris:
Oh shit.  You're right.  I am a jet-lagged idiot.  What a diabolical little bug you found there.   Merged.   Thank you.

garris:
I will add that test to the feature_test script.

pkra:
Thanks so much for taking another look, @garris. I'm glad I could contribute something back.

And my apologies for not providing a minimal example in the first place; I know how annoying half-baked reports are.

PR #813, opened 2018-06-26
re-organizing dependencies, moves HTML report tool supporting files to dev dependency

visionmonster:
I think this should allow you to install backstop as a dependency without requiring projects to also install React libraries.


visionmonster:
Does this structure for dependencies work for you? If it does, do you know when it might be pulled in? Just looking for a time frame to tell my people. Let me know if you have questions. This is a great package.

garris:
@visionmonster Thanks for this.   I think there have been a few iterations of this — with various users running into issues.   Can you explain in detail what problem this should fix and what you’ve done to fix it?  Please site npm docs where necessary and please also detail scenarios where this change may cause different behavior from what we currently see?

visionmonster:
sure, so what I was running into is I am trying to run BackstopJS as a gulp task, however, in the project I am working on I can't be sure other developers will have BackstopJS installed globally. So when you do an `npm install backstopjs --save-dev` this makes BackstopJS a dev dependency. Our project won't ship with Backstop, just like it won't ship with our unit tests.

For example in BackstopJS in order to get the console to log out in color, Chalk is required.  If Chalk wasn't in the "dependencies" list Backstop wouldn't be able to run because the package  would be missing. 

If you want to run just the testing portion of Backstop you don't need React, since React has already been flattened and is hanging out, bundled by web pack, in the backstopjs/compare/src folder

a quick example of the two different package.json files:
```
mkdir test-current
cd test-current
npm init
```
just hit enter until it's done creating a new project.
```
npm install backstopjs --save-dev
```
npm will do it's thing and let you know a few things as well and should finish with something like:
```
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN backstopjs@3.5.2 requires a peer of react-dom@^15.6.1 but none is installed. You must install peer dependencies yourself.
npm WARN react-modal@3.5.1 requires a peer of react-dom@^0.14.0 || ^15.0.0 || ^16 but none is installed. You must install peer dependencies yourself.
npm WARN react-sticky@6.0.3 requires a peer of react-dom@>=15 but none is installed. You must install peer dependencies yourself.
npm WARN ajv-keywords@3.2.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN test@1.0.0 No repository field.
```
if you checkout your test-install/node_modules folder, you would see something like `react` , `react-lifecycles-compat`, etc
but you would still be able to run Backstop as a visual regression testing tool. 
from your test-current folder:
```
node ./node_modules/backstopjs/cli/index.js reference --config=node_modules/backstopjs/test/configs/backstop
node ./node_modules/backstopjs/cli/index.js test --config=node_modules/backstopjs/test/configs/backstop
```
Now lets try out the proposed configuration:
```
mkdir test-proposed
cd test-proposed
npm init
```
again, just hit enter until it's done creating a new project.
```
npm install git+https://github.com/visionmonster/BackstopJS.git --save-dev
```
again npm will do it's thing and let you know a few things as well but this time finishes with something like:
```
npm WARN test-proposed@1.0.0 No description
npm WARN test-proposed@1.0.0 No repository field.

+ backstopjs@3.5.2
```
This time if you look at your test-proposed/node_modules you won't have any react packages.

you can still run backstop from your test-proposed directory:
```
node ./node_modules/backstopjs/cli/index.js reference --config=node_modules/backstopjs/test/configs/backstop
node ./node_modules/backstopjs/cli/index.js test --config=node_modules/backstopjs/test/configs/backstop
```
The difference being if I now run `npm install react --save-dev` from my `test-proposed` directory I can successfully install `+ react@16.4.1` with no conflicts with BackstopJS, since those files are already bundled.

but if try that on test-install I get a warning that says: backstopjs@3.5.2 requires a peer of react@^15.6.1 even though i have react@16.4.1 installed.

For this scenario I'm not doing development work on Backbone, but I am trying to use as part of my suite of testing tools, like eslint and Tape tests.

The proposed package.json would still allow you to develop BackboneJS, you can clone the repo and run npm install and all the react packages will be included, since they are required to run the `"build-and-copy-report-bundle"` task

from the npm docs:
https://docs.npmjs.com/files/package.json#devdependencies

> If someone is planning on downloading and using your module in their program, then they probably don't want or need to download and build the external test or documentation framework that you use
In this case, it's best to map these additional items in a devDependencies object.

some more good reading:
https://stackoverflow.com/questions/22891211/what-is-the-difference-between-save-and-save-dev

It wasn't until I ran into this with an electron project I was working on that I started to understand what was going on.  But I had it in reverse, where everything I had was in devDependencies and those packages don't get bundled in the electron application.

I think the project like https://github.com/jparkerweb/bivariate doesn't have any need for React, since it's entirely command line driven. 

Let me know if you would like any more detail or have any more questions. 
I appreciate you taking the time to follow up.

--Matt

visionmonster:
@garris How do these examples work for you? Do you have any more questions about what is trying to be solved? Thanks!

garris:
Hey Matt — so sorry for the delay!  I’ve been traveling and got really behind on reviews.  Any case,  this is a great write-up!   You should consider making it a blog post!

Thank you very much for the PR and the detailed documentation here.   I am planning the next release for next Friday and this will go out with that.  

Cheers!

Issue #802, opened 2018-06-15
Stop Chrome console logs?


smvilar:
Is there a way to prevent the logs from the browser console to show up in the `test` output?
I looked at the Chromium command line flags but I didn't find the one.

garris:
There are no switches for this— but you can probably unbind the handler for log events in an onBefore script.  Implementation would depend on the engine option.

verbruggenalex:
This is the only reference I can find. The puppet engine also uses chromy?
https://github.com/garris/BackstopJS/blob/46a93e039f56fe8a8e0d0f21a50c4265f08459b4/capture/backstopTools.js#L8-L21

I'm not very proficient at javascript. Could someone help me with a small implementation to unbind these log messages? Eventually I think that there should be some sort of log level setting made available. The messages "Browser Console Log 0: JSHandle:BackstopTools have been installed." and "x Close Browser" feel like pollution to me.

<a href="https://ibb.co/b7n0Td"><img src="https://preview.ibb.co/bEVyEJ/console_output_pollution.png" alt="console_output_pollution" border="0"></a>

garris:
Ok.  I just looked at this for you.

Here is the puppeteer documentation which shows how to do it... https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-page

so the ideal way to address this would be for you to add this in the onBefore script...
```
module.exports = async (page, scenario, vp) => {
  page.removeListener('console', refactoredConsoleMethodHere)
  await require('./loadCookies')(page, scenario);
};

```
Unfortunately --this wont work.  The inline anonymous function would need to be slightly refactored per the puppeteer docs for it to correctly work.   It would be awesome if you could do this an submit a PR.   


^  If that is not something you're comfortable with you could probably solve your problem with something totally crude like this...

![image](https://user-images.githubusercontent.com/447033/42787409-420fcc3c-890f-11e8-9a6c-2e92f5e4309f.png)

It would be way way better to do the first suggestion.  Don't tell anybody I told you to kill the console like this.  It's utter trash -- but it will kill your output. 😉

verbruggenalex:
Thank you for the swift answer. I created something similar but still nasty:

<details><summary>onBefore.js</summary>

```javascript
(function (original) {
    console.enableLogging = function () {
        console.log = original;
    };
    console.disableLogging = function () {
        console.log = function () {};
    };
})(console.log);

module.exports = async (page, scenario, vp) => {
[...]

PR #822, opened 2018-07-17
Proposal for using level logging (in runPuppet.js)

verbruggenalex:
Just want some feedback. I know this is not acceptable just yet. But it's 2:30 in the morning for me :).
This originated from https://github.com/garris/BackstopJS/issues/802

If this is a direction you are willing to go. I can put some effort in getting to know the runners and loggers.

### DIFF
--- a/core/util/runPuppet.js
+++ b/core/util/runPuppet.js
@@ -67,7 +67,7 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
   page.setDefaultNavigationTimeout(engineTools.getEngineOption(config, 'waitTimeout', TEST_TIMEOUT));
 
   if (isReference) {
-    console.log(chalk.blue('CREATING NEW REFERENCE FILE'));
+    console.info(chalk.blue('CREATING NEW REFERENCE FILE'));
   }
 
   // --- set up console output and ready event ---
@@ -82,7 +82,7 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
   page.on('console', msg => {
     for (let i = 0; i < msg.args().length; ++i) {
       const line = msg.args()[i];
-      console.log(`Browser Console Log ${i}: ${line}`);
+      console.debug(`Browser Console Log ${i}: ${line}`);
       if (readyEvent && new RegExp(readyEvent).test(line)) {
         readyResolve();
       }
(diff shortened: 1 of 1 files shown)


garris:
@verbruggenalex thank you very much for posting this proposal!   

I think I maybe did not give a clear direction before so please allow me to clarify.  The intention here is to continue to use `console.log` exclusively as an output channel.  (FWIW: In node, `console.debug` is just an alias for `console.log`).   Instead of mapping to different console methods/aliases I think it is more our job **1)** to determine the nature of the output (I appreciate and very much agree with your proposed classifications) and **2)** map that to a logging level so that the logging points you have identified can be enabled/disabled per a global `logLevel` property in the config or `logLevel=<logLevel>` flag on the command line.

`logLevel` should be something like this -- implemented in an enum-like interface
```
0 OFF (or if `logLevel  === FALSE`)
1 DEBUG (or if  `logLevel | debug  === true`)
2 INFO (or if  `debug  === FALSE` for backward compatibility)
3 WARN
4 ERROR
5 FATAL
6 OFF
```

We should also create chalk styles similar to the approach below and map these names to our logging level constants...
image

Currently in our configs we have a `debug` property.  This should be deprecated (and removed from the docs).  This should be replaced with something like, `logLevel` (implemented as an alias of `debug` for permanent backward compatibility.  This log level should take either the integer or string values from `logLevel` above.

Again,  I know this is a lot of work -- so no worries if this is too much to implement (but of course I gladly take a PR if you are so inclined ☺️).   I felt that I should at least be specific as to the ideal place we'd like to be with the logging feature. I think the above spec is closer to what developers would expect from a mature application. 

Cheers.

verbruggenalex:
Hi @garris , thanks again for the detailed explanation. I don't see many projects giving such great support.

I fully agree that good logging is essential for developers. Especially when they are using CI. I will do my best to put some efforts into following your directions. But I might make a pull request first for a simplified "progress" formatter like behat is providing. So essentially all you do is output a single character per scenario executed.

```
    [behat] Executing command: /test/toolkit/tests/vendor/behat/behat/bin/behat --config="/test/toolkit/tests/behat.cwt.yml" --strict
...................................................................... 70
...................................................................... 140
.F----F-.F----......F-----......F-----......F-----......F-----......F- 210
----......F-----......F-----...F----.........F--------------.........F 280
```

After that it will be much easier to categorize and format the pretty output of the logs. My thinking is to backengineer that output from the generated data set in config.json to output the data in a way that will become "templatable". Any output not related to the result set itself should then be considered debug output.

I'm more proficient at PHP and at symfony/console. So I will probably do the backengineering of the result set into console output in that. With this it will become much easier to analyse what needs to be done to refactor the logger system to become more configurable by having all the possible raw data properties sent to the logger to be formatted only before outputting.

But like you said. This may be a lot of work. A progress formatter will probably have less impact on the current structure. If you agree with this approach I'd be more than happy to create a pull request for the progress formatter.

garris:
@verbruggenalex Thank you for the input.   I am not familiar with behat -- it seems like a generally good idea, but it would need to be something I'd need to learn more about.   I am not clear how that would integrate from an API standpoint.

fuhlig:
This PR & issue is stale while I very like the idea and concept.

Is there any progress or help needed?

PR #825, opened 2018-07-23
Lazy load report images

justinph:
This change makes viewing the generated HTML reports more efficient and user-friendly by not loading test images that are not yet in the viewport. This makes it far more practical to view large reports quickly and not have the page consume large amounts of bandwidth. 

This change adds and uses [react-visibility-sensor](https://github.com/joshwnj/react-visibility-sensor) to handle the visibility detection. This is the library used on www.nytimes.com to lazily load images. 

Note: please ignore the extra early commits in my PR from some earlier work. I assume this will get squash-merged so they won't matter.


garris:
@justinph     Thanks for the PR -- I have been wanting to see this for a long time!   Finally! 😄    I will check this before pushing to NPM -- but before I do I just want to make sure you checked this against `npm run smoke-test`.  And also did you happen to test this with different display states? (eg... 
 viewing all pass, all fail, changing search criteria and excluding diff, and/or reference images in report via the control to the right of the search window)    Please let me know.    Thanks super a lot!

justinph:
@garris: Yup, it looked all good from the smoke tests. Also ran it against our own configs and it works well, no discernible issues.

PR #831, opened 2018-08-07
Fix test command when using docker

BBlackwo:
- `backstop test --docker` was actually running `reference` instead. Fixed that and did a bit of refactoring so there's less copied and pasted code. 
- If you have spaces in a command it now adds quotes around it when passing through to Docker. E.g. `backstop test --docker --filter="some test"`.
- Now `--docker` will work on window as I'm using `${process.cwd()}` instead of `$(pwd)` in the docker command.
- Also added a `CONTRIBUTING.md` guide. It's a start. I couldn't figure out what commands to run and such to test my changes. Had to dig through the `package.json > scripts`. I think it could help others too.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,9 @@
+# Contributing to this project
+
+## Test
+
+```sh
+npm run genConfig
+npm run reference
+npm run test
+```
--- a/core/command/reference.js
+++ b/core/command/reference.js
@@ -1,30 +1,12 @@
-var createBitmaps = require('../util/createBitmaps');
-var fs = require('../util/fs');
-var logger = require('../util/logger')('clean');
+const createBitmaps = require('../util/createBitmaps');
+const fs = require('../util/fs');
+const logger = require('../util/logger')('clean');
+const { shouldRunDocker, runDocker } = require('../util/runDocker');
 
 module.exports = {
   execute: function (config) {
-    if (config.args.docker) {
-        const passAlongArgs = process.argv
-                    .slice(3)
-                    .join(' ')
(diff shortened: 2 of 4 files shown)


BBlackwo:
Any chance you can merge this one in @garris? It's been sitting around for almost a week now.

brendonbarreto:
@BBlackwo I'm having some trouble running with docker. 

```js
if (config.args.docker) {
    const passAlongArgs = process.argv
                  .slice(3)
                  .join('" "') // in case of spaces in a command
                  .replace(/--docker/, '--moby');
```

And in my case, i don't want to use `--docker`, i want set this on my configuration file.
I know that this have nothing to do with your code, but maybe you could give the user the choice of running with `--docker` or set it on configuration file.

What do you think?

garris:
@BBlackwo Just occurred to me -- you're adding quotes to docker args in this PR...   What happens if the user included these quotes in the original argument?  👉 e.g. `--filter="myScenarioName"`

BBlackwo:
@brendonbarreto being able to set it on the config file sounds reasonable. Maybe you can submit a PR for that :)

@garris thanks for the merge. Sorry I didn't see that section in the readme. Maybe you could add it to the table of contents at the top?  Happy if you want to remove the CONTRIBUTING doc - doesn't make sense to have it in 2 seperate places. 

Also with the quotes, before my change if you ran --filter="My Scenario Name" the quotes would be stripped off when running the docker command. My change always puts them in case of spaces. The `process.argv` array didn't include the quotes in the arg, but just treated each arg as a string.

BBlackwo:
Added it also to the table of contents in #838

Issue #834, opened 2018-08-10
Docker issue with local install


brendonbarreto:
A have a different setup for backstopjs. Basically I have a node file where I manually invoke `backstop`

```js
backstop('test', config)
```

This node file, can be executed with other parameters like `url`

`node tests_vrt/backstop.test.js localhost:8001`

How can I run with docker in this case?

garris:
Does it work when you pass `{docker: true, ... }`  in the config?

brendonbarreto:
No, I tried `docker: true` and `args: { docker: true }`.
And I don't know if this was supposed to work, because of this part of the `core/command/test.js`
```js
const passAlongArgs = process.argv
	                .slice(3)
	                .join(' ')
	                .replace(/--docker/, '--moby')
```

garris:
Hmmm,  I thought the docker argument would have been passed from the config obj...   anyway, i am happy to review a PR if you want to add this.   Cheers.

garris:
@brendonbarreto  I just looked at my notes from the docket release.  Adding `{docker: true, ... }` as part of the config is equivalent to using the `-- docker` flag on the cli.

This worked as of the last release.  Can you please try this again?

brendonbarreto:
Yes, that worked for me. Thanks!

garris:
Thanks for reporting! 👍

PR #837, opened 2018-08-13
Docker support on configuration file

brendonbarreto:
With this PR, user can run with docker by adding `docker: true` on configuration file. I don't know if this is the best solution but it worked for me. 

@garris The only difference between running with `--docker` and `docker: true` is this:

```js
const passAlongArgs = process.argv
                  .slice(3)
                  .join('" "') // in case of spaces in a command
                  .replace(/--docker/, '--moby');
```
The `DOCKER_COMMAND` will be without `--moby`. Should I be worried about this?

fixes https://github.com/garris/BackstopJS/issues/834


brendonbarreto:
Implemented by @garris in https://github.com/garris/BackstopJS/commit/af804531c2208359dd071f13e1f7b9348ee3e9c9

PR #840, opened 2018-08-17
Docker volumes are enough

caiofbpa:
In our use case using a volume instead of a mount is much better:

- `mount` is available only from Docker version 17.06, while `volume` is available since ancient Docker times
- We don't need to assert that `/src` exists on the container prior to binding, which is the main benefit of `mount`
- `volume`'s syntax is more straightforward

I also took the liberty of removing the `-i` flag since we don't need interactivity during tests, and when run by a non-interactive terminal (like inside an IDE), it throws `The input device is not a TTY`.

### DIFF
--- a/core/util/runDocker.js
+++ b/core/util/runDocker.js
@@ -17,7 +17,7 @@ module.exports.runDocker = (config, backstopCommand) => {
       configArgs = configArgs.replace(/--docker/, '--moby');
     }
 
-    const DOCKER_COMMAND = `docker run --rm -t -v /src:${process.cwd()} backstopjs/backstopjs ${backstopCommand}${configArgs} "${passAlongArgs}"`;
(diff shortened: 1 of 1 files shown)


gabegorelick:
While this is somewhat better than using bind mounts, it still won't work when docker is running on a remote host (https://github.com/garris/BackstopJS/issues/858). I think an even better, albeit more complicated solution, would be to manually `docker cp` into the volume.

garris:
@gabegorelick I think your suggestion would enable "local" usecases -- correct?  E.g. when URL prop points to a file.

I guess downside is if there is a ridiculous number of files in there it would be pretty clunky to cp everything all the time.

gabegorelick:
> I guess downside is if there is a ridiculous number of files in there it would be pretty clunky to cp everything all the time.

Yes, that is a very good point. On my code base `docker cp` does take a long time. So it may not be worth it if you're running Docker locally and thus don't need it.

Maybe we could only fallback to the `docker cp` method if `process.env.DOCKER_HOST` is set?

caiofbpa:
I'll close this as I assume the issue will be resolved in another fashion

Issue #842, opened 2018-08-23
Smoke tests are flaky because of CSS transition


KenCoder:
I'm separating this issue out of #747 for clarity. This CSS line:

```css
.lemurInYourFace {
  transition: transform 750ms ease, opacity 750ms ease, visibility 800ms ease;
```

Means that any tests which grabs its screenshot in the first 750ms of the page will get a different result. If I repeatedly run the smoke test on my Macbook, I can get anywhere from 0 to 8 failures. 

For websites with a transition like this, the right way to handle it would be to detect the transitionend event, and not grab any screenshots until that event is triggered. That could be a new backstop feature, or it could be something users do manually and then use the readyEvent.

You probably don't want to comment out the tests, as suggested in the PR, because you'd have to comment out lots of tests - any tests that doesn't wait for the transition in theory could be affected, in practice only fast tests have the problem. 

If you remove this transition, then I get 100% consistent smoke test results. I'll submit a 1 line PR for this.

KenCoder:
Here's an example:
<img width="1351" alt="screen shot 2018-08-23 at 11 19 26 am" src="https://user-images.githubusercontent.com/403709/44535288-a3b43880-a6c7-11e8-9204-3473fbffe873.png">

KenCoder:
To clarify, the transition means that any change to the lemur face happens over 750ms. So any test that changes the face could be flaky because that change happens over 750ms. For example, the reference images currently on master for the hide selectors test always have the face in the image, even though the face is in a <p> element which should be hidden by the test.

PR #843, opened 2018-08-23
Address #842 by removing CSS transition

KenCoder:


gabegorelick:
@garris Any reason not to merge this?

garris:
Yes.  This is a real-life use case.  One can't just change product design because it's inconvenient to test.   There are plenty of ways to approach this.

gabegorelick:
I'm not sure I follow? Is the CSS transition on the lemur that integral that's it's worth not having smoke tests?

I agree that it would be nice to get smoke tests working with transitions, but if the alternative is no smoke tests, to me the choice is obvious.

Perhaps we can merge this, get smoke tests working, then work on bringing the transition back?

PR #862, opened 2018-09-24
Revert c02a163

gabegorelick:
Not sure if https://github.com/garris/BackstopJS/commit/c02a1638f266d9e154a49541315af2f80312d265 was an intentional commit, but it has a lot of lint and syntax errors. The Docker image it generated is also broken:
```
Status: Downloaded newer image for backstopjs/backstopjs:latest
/usr/local/lib/node_modules/backstopjs/core/command/remote.js:13
    for await (const data of child.stdout) {
        ^^^^^

SyntaxError: Unexpected reserved word
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:617:28)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at requireCommand (/usr/local/lib/node_modules/backstopjs/core/command/index.js:44:26)
```  

See https://github.com/garris/BackstopJS/issues/861.

PR #864, opened 2018-09-24
Tech debt cleanup of UI + Plumbing for Remote feature & "Diverged" image diffing

garris:
This PR replaces c02a163 (#862) which was admittedly wasn't completely baked. Sorry.

Anyway this PR...
- cleans up some tech debt where UI image scrubbing and `backstop init` command were originally implemented as hacks.
- Add the `backstop remote` command.   Running `backstop remote [--config=<path>]` in a separate terminal launches a simple web server such that reports can be shared live with all your buddies.  The `backstop openReport` command now checks for a server running on localhost and will automatically choose the web hosted version if found.
- added "Diverged" image diff library -- in development here... https://github.com/garris/diverged -- so we can eventually have our diffs be more awesomer.


garris:
@gabegorelick  -- would you please take a look at this when you get a chance.  I have fixed all the linting issues.   I would like to push this to master to resolve some UI issues.   I will bump the version prior.   Thank you.

garris:
merged offline

PR #868, opened 2018-09-26
Better Docker error handling

gabegorelick:
- Handle `error` event
- Reject with an error so we preserve stack trace

### DIFF
--- a/core/util/runDocker.js
+++ b/core/util/runDocker.js
@@ -1,3 +1,4 @@
+const { spawn } = require('child_process');
 const version = require('../../package').version;
 
 module.exports.shouldRunDocker = (config) => config.args.docker;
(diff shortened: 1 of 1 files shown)


core/util/runDocker.js
```diff
@@ -20,16 +21,16 @@ module.exports.runDocker = (config, backstopCommand) => {
     }
 
     const DOCKER_COMMAND = `docker run --rm -it --mount type=bind,source="${process.cwd()}",target=/src backstopjs/backstopjs:${version} ${backstopCommand}${configArgs} "${passAlongArgs}"`;
-    const { spawn } = require('child_process');
```
gabegorelick: Moving this up since requiring a module is sync the first time it's called..

garris:
Docker build failed tho — looks like maybe due to a weird network or service issue.  Retrying.

PR #869, opened 2018-09-26
Clean up linting and run it in Travis

gabegorelick:
Running linting in Travis gives contributors and maintainers more confidence in PRs.

Most of the changes are from removing unused variables to get linting to pass.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,0 @@
-bower_components
--- a/compare/src/components/atoms/ImageScrubber.js
+++ b/compare/src/components/atoms/ImageScrubber.js
@@ -1,12 +1,8 @@
 import React from 'react';
-import { connect } from 'react-redux';
 import styled from 'styled-components';
 import TwentyTwenty from 'backstop-twentytwenty';
 import { colors, fonts, shadows } from '../../styles';
 import diverged from 'diverged';
-
-const BASE64_PNG_STUB =
-  'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
 
 const ScrubberViewBtn = styled.button`
   margin: 1em;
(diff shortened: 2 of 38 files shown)


.eslintignore
```diff
@@ -1 +0,0 @@
-bower_components
```
gabegorelick: This isn't used since an explicit list of files to lint is passed to the `lint` npm script.

compare/src/components/atoms/ImageScrubber.js
```diff
@@ -99,7 +95,7 @@ export default class ImageScrubber extends React.Component {
 
     const scrubberTestImageSlug = this.props[testImageType];
 
-    function getDiverged (arg) {
+    function getDiverged (arg) { // eslint-disable-line
```
gabegorelick: This function is unused but I assume it will be used as part of the new diverged work.

compare/src/components/ecosystems/ScrubberModal.js
```diff
@@ -65,10 +63,6 @@ const customStyles = {
 };
 
 class ScrubberModal extends React.Component {
-  constructor (props) {
```
gabegorelick: There's a `no-useless-constructor` rule that flags this.

package.json
```diff
@@ -6,10 +6,7 @@
     "backstop": "./cli/index.js"
   },
   "scripts": {
-    "lint": "npm run lint-web -- --fix && npm run lint-node -- --fix",
-    "lint-ci": "npm run lint-web && npm run lint-node",
-    "lint-web": "eslint --env browser compare/*.js compare/src/**/*.js",
-    "lint-node": "eslint --env node core/*.js core/**/*.js cli/*.js cli/**/*.js capture/*.js capture/**/*.js",
+    "lint": "eslint '*.js' 'compare/*.js' 'compare/src/**/*.js' 'core/**/*.js' 'cli/**/*.js' 'capture/**/*.js'",
```
gabegorelick: Combining all these scripts into one since what env to use is controlled by `.eslintrc` files.

gabegorelick: After this is merged we should probably enabled linting for test directory too.

compare/src/components/atoms/ImageScrubber.js
garris: Correct

PR #871, opened 2018-09-26
Allow running `backstop --config file test --docker`

### LINKED ISSUE #867: backstop --docker ignores options specified before command
I would expect the following to work:
```bash
backstop --config file test --docker
```
However, due to https://github.com/garris/BackstopJS/blob/6634b13bd7b0aa9a61697d410f3aa9b3ae96c740/core/util/runDocker.js#L7-L8, what actually gets executed is `docker run ... test "file" "test" "--moby"`.

While that may seem like a convoluted way to pass arguments, it's common when using npm scripts to compose commands. E.g. I have something like the following in my package.json `scripts` sections:
```json
{
  "backstop": "backstop --config ./test/backstop.js",
  "backstop:test": "npm run backstop test --docker",
}
```

gabegorelick:
Fixes #867

### DIFF
--- a/core/util/runDocker.js
+++ b/core/util/runDocker.js
@@ -4,10 +4,13 @@ module.exports.shouldRunDocker = (config) => config.args.docker;
 
 module.exports.runDocker = (config, backstopCommand) => {
   if (config.args.docker) {
-    const passAlongArgs = process.argv
-                  .slice(3)
-                  .join('" "') // in case of spaces in a command
-                  .replace(/--docker/, '--moby');
+    // 0th element is node, 1st is backstop, 2nd may be command or an option like --config
+    const args = process.argv.slice(2);
+    args.splice(args.indexOf(backstopCommand), 1);
+
+    const passAlongArgs = args
+      .join('" "') // in case of spaces in a command
+      .replace(/--docker/, '--moby');
 
     // When calling BackstopJS from node config props will be overridden by the passed config object. e.g. backstop('test', {thisProp:'will be passed to config.args'})
     // NOTE: passing config file name is supported -- passing actual config data is not supported.


gabegorelick:
I've tested this locally, but not sure how to construct a unit test for this.

garris:
Unit test would be great.  I think you’d want to stub the spawn method with a sinon wrapper and just test that various inputs produce the correct output command.

garris:
Sinon has a sandbox feature which should help with pollution.   All tests would need to run cleanup too.

gabegorelick:
I’m familIar with sandboxes and am using them. I think the issue is that
backstop calls process.exit thereby killing mocha. Ideally the tests would
fork a subprocess and run backstop in that. But that’s a larger change.

garris:
Thanks for all your help here Gabe— I really appreciate it.   I like that you split up your changes into small bite sized enhancements.  Much easier to review and push out.  

Unit test are currently disabled / ignored in Travis so it’s fine with me to skip that.  

Mostly, I run the sanity-test, the backstop_featues test and the `run from node` test to determine if it’s safe enough to publish.   I have a super challenging day job and don’t have a lot of time to put into enhancements — but if you want to keep improving the test coverage with very small PRs I am happy to support that.  It will just take some time — most of my reviews are done on the train to work via iPhone 😉.   Cheers!

gabegorelick:
> Just want to make sure this was tested with runFromNode

Not sure I follow how to test with `runFromNode`. Do you mean just verifying that the following works?
```js
backstop('test', {
  docker: true,
  config: 'backstop_alt'
});
```

If so, it does :)

Also, I appreciate all your work! My hope with a lot of the CI changes is to make it easier for you to review changes, which would allow more outside contributions and be more confidence in new features (I don't want my build to fail again from upstream changes 😉). 

Thanks again.

garris:
@gabegorelick     Boom!  https://github.com/garris/BackstopJS/releases/tag/v3.7.0

PR #872, opened 2018-09-26
Run all tests in `unit-test` script

gabegorelick:
Not just the ones in `test/cli`.


gabegorelick:
Yep. Want to land https://github.com/garris/BackstopJS/pull/873 first so mocha will actually fail. Then I'll try to figure out the test failures.

PR #873, opened 2018-09-26
Don't override mocha's exit code

### LINKED ISSUE #750: Travis build passes even if unit tests fail
There are currently two unit tests in tests/cli, and one of them (usage_spec.js) is failing. But the travis build is succeeding. I investigated this, and for some reason mocha is return a 0 exit code even though a test is failing. If I test with only usage_spec.js, mocha correctly returns 1. 

I tried a few versions of mocha but couldn't find any that fixed it. 

Maybe I could try a switch to jest? We use it a lot and haven't had this kind of problem.

gabegorelick:
Fixes #750

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -43,8 +43,8 @@ if (!commandName) {
   /*
    * Wait for the stdout buffer to drain.
    */
-  process.on('exit', function () {
-    process.exit(exitCode);
+  process.on('exit', function (code) {
+    process.exit(code || exitCode);
   });
 
   process.on('uncaughtException', function (err) {


gabegorelick:
Sorry, I realize there may have been some confusion here. By returning 1 on tests failures, I meant _mocha_ tests, not `backstop test` tests. Only mocha test failures will return 1. `backstop test` will still return 0, unless there's some catastrophic error, even if there are scenarios that did not match the references.

In summary, this change is about backstop's test suite returning an error code when it fails. Nothing else should be affected.

PR #876, opened 2018-09-27
Don't force exit on makeConfig errors

gabegorelick:
Instead, allow error to bubble up. End result is the same when running via CLI, but when running in another process you don't want to force exit.

### DIFF
--- a/core/util/makeConfig.js
+++ b/core/util/makeConfig.js
@@ -43,15 +43,10 @@ function loadProjectConfig (command, options, config) {
       }
       userConfig = options.config;
     } else if (config.backstopConfigFileName) {
-      try {
-        // Remove from cache config content
-        +delete require.cache[require.resolve(config.backstopConfigFileName)];
-        console.log('Loading config: ', config.backstopConfigFileName, '\n');
-        userConfig = require(config.backstopConfigFileName);
-      } catch (e) {
-        console.error('Error ' + e);
-        process.exit(1);
-      }
+      // Remove from cache config content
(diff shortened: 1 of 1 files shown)


core/util/makeConfig.js
```diff
@@ -43,15 +43,10 @@ function loadProjectConfig (command, options, config) {
       }
       userConfig = options.config;
     } else if (config.backstopConfigFileName) {
-      try {
-        // Remove from cache config content
-        +delete require.cache[require.resolve(config.backstopConfigFileName)];
```
gabegorelick: Does the + before `delete` do anything?

core/util/makeConfig.js
```diff
@@ -43,15 +43,10 @@ function loadProjectConfig (command, options, config) {
...
-        console.log('Loading config: ', config.backstopConfigFileName, '\n');
-        userConfig = require(config.backstopConfigFileName);
-      } catch (e) {
-        console.error('Error ' + e);
```
gabegorelick: BTW, better way to log an error is `console.error(e.stack)`. Since the stack is already preceded by the error's `name`, you don't want to print `Error Error: ...`.

core/util/makeConfig.js
garris: I don't think so.  Looks like a typo. 🙄

garris:
I preferred the less verbose error message -- but I guess ending the process abruptly there can be problematic.  Alrighty.

PR #878, opened 2018-10-01
Bump puppeteer

brendonbarreto:
Puppeteer was in a version over 6 months ago. I was having a problem that the click on the element was not working properly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -93,9 +93,9 @@
       }
     },
     "agent-base": {
-      "version": "4.2.0",
-      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.2.0.tgz",
-      "integrity": "sha512-c+R/U5X+2zz2+UCrCFv6odQzJdoqI+YecuhnAJLa1zYaMc13zPfwMwZrr91Pd1DYNo/yPRbiM4WVf9whgwFsIg==",
+      "version": "4.2.1",
+      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.2.1.tgz",
+      "integrity": "sha512-JVwXMr9nHYTUXsBFKUqhJwvlcYU/blreOEUkhNR2eXZIvwd+c+o5V4MgDPKWnMS/56awN3TRzIP+KoPn+roQtg==",
       "requires": {
         "es6-promisify": "^5.0.0"
       }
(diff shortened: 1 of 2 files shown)


garris:
Thank you!

garris:
@brendonbarreto   Sorry -- I had to revert your bump.   Please take a look at the attached `backstop_features` test.   This version of puppeteer is clearly rendering differently.   Can you please review (you can run tests locally using the previous version and the latest version and you will likely see the same difference)

I do think it is a good idea to bump puppeteer -- but we need to notify other developers what changes (failures) they should expect.

- Can you please describe the differences in rendering behavior and what situations cause the differences?
- What should other users do if they upgrade backstop and their tests have broken?  
- Are we totally sure this version of puppeteer is the most stable version to use?


Thank you!

brendonbarreto:
Sorry about that.

These positioning differences are happening in tests that have `selectors`. They are happening because in the version that was, the puppeteer forced a scroll until near the element. In the new version, this scroll does not happen. This way the content is in a different position compared to the background. I have not gone too far on how to circumvent this behavior of the puppeteer.

I could only check the real problem by starting the browser in `slowMo`
https://github.com/GoogleChrome/puppeteer/blob/v1.8.0/docs/api.md#puppeteerlaunchoptions

garris:
I suspect these are not "problems".   Perhaps puppeteer has settled on a new expected behavior for capturing "selectors"?   I think we just want to understand this change and then make backstop users aware of the change by pointing them to the correct documentation.   It would be great if you could investigate this -- maybe just look at puppeteer revision history and find where they made the change?   We can add a link to this in our documentation and then reapply your PR.   I am happy to do this quickly since I expect puppeteer has added many improvements since 1.2.x.

brendonbarreto:
I think you're right. I'll be working on this later today

brendonbarreto:
We are currently using `1.2.0-next.1523485686787`. The next version is `1.2.0-next.1523546977615`. This next version, as the last one `1.8.0`, does not scroll to print selector.

This scroll was never a stable behavior. `1.2.0` and `1.3.0` works like `1.8.0`. 

Just wanted to share some discoveries.

brendonbarreto:
I could not find the exact PR or commit that bring this scroll behavior. It could also be a chromium revision.

And puppeteer just release 1.9
https://github.com/GoogleChrome/puppeteer/releases/tag/v1.9.0

garris:
Ok.  it sounds like you did good research -- perhaps it is actually chrome tools that made the change.  In any case yes, lets bump to the latest version.   How should we notify users that we expect some of their tests will break?   Maybe in troubleshooting section of readme?   Any other ideas?

garris:
We should definitely at least bump our minor version.

brendonbarreto:
As you suggested, I think we could use troubleshooting section. Maybe something like this?

### Puppeteer bump

In the version we were in, puppeteer forced a scroll to take screenshot, when running a scenario with `selectors` defined. This can cause difference in the images captured before and after the bump. This issue will probably happen if in your site the backgroud is fixed to the viewport, or it uses the parallax effect.

In this case simply approve the changes, since now puppeteer takes screenshot without scrolling.

**NOTE:** english is not my primary language, so sorry for grammar issues.

PR #888, opened 2018-10-08
Bump dependencies: eslint

brendonbarreto:
## Reason behind some changes

- `fs.exists` -> `fs.existsSync`: `fs.exists` is deprecated: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback
- spacing inside braces: https://eslint.org/docs/rules/object-curly-spacing
- https://eslint.org/docs/rules/no-mixed-operators
- https://eslint.org/docs/rules/prefer-promise-reject-errors
- `assert.equal` is deprecated: https://nodejs.org/api/assert.html#assert_assert_equal_actual_expected_message

## Vulnerabilities
Bumping ESLint we remove 3 low vulnerabilities

| Before             |  After |
:-------------------------:|:-------------------------:
  |


brendonbarreto:
So can I delete `compare/output` from git, since It's generated?

And then add `compare/output/*` to gitignore?

garris:
@brendonbarreto   We actually need `compare/output` in git as it is generated.  This resource is copied into `html_report` during runtime.  Is it good enough to simply whitelist `index_bundle.js` anywhere in the repo?

brendonbarreto:
Since ESLint as you said, is already ignoring `index_bundle.js`, this is ready to merge. I just got confused about what to do with this file 😄

garris:
Merged!

Issue #880, opened 2018-10-03
Some security issues


garris:
Would love some help bumping our libs!

Security scan is showing the following...

![screen shot 2018-10-03 at 8 52 13 am](https://user-images.githubusercontent.com/447033/46422649-aa33c800-c6e9-11e8-92d9-11f2628f84a3.png)

garris:
WOW -- thank you @brendonbarreto for all your hard work on this!!!   Just one p3 left.  Cheers! 🌈

![image](https://user-images.githubusercontent.com/447033/46688774-11062500-cbb3-11e8-9350-b1202bdde6a5.png)

garris:
![image](https://user-images.githubusercontent.com/447033/46705399-38291a80-cbe4-11e8-96f7-2cf00c2896a2.png)

Issue #866, opened 2018-09-25
1 low severity vulnerability found by npm audit


GWShark0:
#### Using:
- BackstopJS: `v3.5.14`
- Node: `v8.12.0`
- npm: `v6.4.1`

When installing `backstopjs` in a Node project using npm 6+, I encounter the following security warning emitted from npm's [auditing tool](https://blog.npmjs.org/post/173719309445/npm-audit-identify-and-fix-insecure):

```
                       === npm audit security report ===

# Run  npm install --save-dev backstopjs@3.6.1  to resolve 1 vulnerability
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ lodash                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ backstopjs [dev]                                             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ backstopjs > junitwriter > xmlbuilder > lodash               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/577                       │
└───────────────┴──────────────────────────────────────────────────────────────┘
```
https://nodesecurity.io/advisories/577

#### Problem:

BackstopJS relies on [`"junitwriter": "~0.3.1"`](https://github.com/garris/BackstopJS/blob/6634b13/package.json#L85) which relies on [`"xmlbuilder": "2.6.2"`](https://github.com/schorfES/node-junitwriter/blob/master/package.json#L27) which relies on [`"lodash": "~3.5.0"`](https://github.com/oozcitak/xmlbuilder-js/blob/v2.6.2/package.json#L25). The [`xmlbuilder`](https://www.npmjs.com/package/xmlbuilder) package is currently at `v10.0.0` and the last publish on [`junitwriter`](https://www.npmjs.com/package/junitwriter) was "3 years ago".

#### Recommendation:

Either the `junitwriter` package should be updated (unlikely given the maintenance history of the package) or BackstopJS should consider using a more recently maintained package for writing JUnit XML reports (possibly [`davidparsson/junit-report-builder`](https://github.com/davidparsson/junit-report-builder)?)

garris:
Thanks for the report!  Would be great if Someone could create a test for the junit writer test and then upgrade our library.   Cheers

PR #891, opened 2018-10-09
 Bump dependencies: fix high severity vulnerabilities

brendonbarreto:
This vulnerabilities come from `phantomjs` and `chromy`. There's a discussion about removing this dependencies, so I think that for now, just getting rid of this vulnerabilities is okay. 

To do this I executed `npm update sshpk --depth 5`. 
We have only a low severity vulnerability remaining: https://github.com/garris/BackstopJS/issues/866.

related #880


garris:
Looks good!

Issue #845, opened 2018-08-25
Localhost unreachable in Docker mode


yebrahim:
Things run fine locally, but doing `backstop test --docker` fails with errors trying to reach my server running on my host. I'm using a Mac for my primary machine, but I'm also seeing similar issues on my Ubuntu desktop. Here are different values I tried for localhost:

|       url                            |      error           |
|-----------------------|---------------|
|http://localhost:3000   |     `ENGINE ERROR: net::ERR_CONNECTION_REFUSED at http://localhost:3000`|
| http://127.0.0.1:3000 | `ENGINE ERROR: net::ERR_CONNECTION_REFUSED at http://127.0.0.1:3000`|
|http://10.0.0.215:3000 (my actual IP)| Works!|
|http://host.docker.internal:3000/| `ENGINE ERROR: net::ERR_NAME_NOT_RESOLVED at http://host.docker.internal:3000/`|

As you can see, the only way I can get it to work is by looking up my IP and embedding it in the scenario url, which isn't portable.

I'm running a simple hello world server using the `hello-world-server` npm package.

Any ideas what's going on?

yebrahim:
It's not working on my Mac. :(
Is there a way to pass arguments to the Docker command? Maybe sharing the host network (`--net=host`) would help?

garris:
Anything you put on the command line will get passed through as parameters.  Good luck.🍀

garris:
Thanks for posting the resolution— cheers!

andyrichardson:
I've had a similar issue on Linux when using `127.0.0.1` as the hostname. 

Exposing the container to the host network has resolved the issue. PR incoming.

luke-hill:
Were there ever any updates to this?

Running `ifconfig` on my Ubuntu VM then altering the backstop config file isn't great.

Is there a way we can have an override file that is ignored by source control (A la docker-compose.override.yml), or something else?

I'm aware this is > 2 years old. But I'm hitting this issue now, and only the hacks contained herein work


cc/ @andyrichardson / @garris

garris:
If I am remembering correctly from 3 years ago, there was no alias for Linux at that time.   I would check docker docs.   If this is still the case then your ifconfig solution might just be a solid solution.

If possible, it would be great if you could post a gist of that solution as it may help others.

Cheers.

luke-hill:
[...]

PR #898, opened 2018-10-19
Run contained instance on host network

andyrichardson:
Fix for Linux when `127.0.0.1` connections are refused.

Relating to #845.

### DIFF
--- a/core/util/runDocker.js
+++ b/core/util/runDocker.js
@@ -23,7 +23,7 @@ module.exports.runDocker = (config, backstopCommand) => {
       configArgs = configArgs.replace(/--docker/, '--moby');
     }
 
-    const DOCKER_COMMAND = `docker run --rm -it --mount type=bind,source="${process.cwd()}",target=/src backstopjs/backstopjs:${version} ${backstopCommand}${configArgs} "${passAlongArgs}"`;
+    const DOCKER_COMMAND = `docker run --rm -it --net="host" --mount type=bind,source="${process.cwd()}",target=/src backstopjs/backstopjs:${version} ${backstopCommand}${configArgs} "${passAlongArgs}"`;
     console.log('Delegating command to Docker...', DOCKER_COMMAND);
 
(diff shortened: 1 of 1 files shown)


garris:
@andyrichardson Would you mind documenting what this change does & how it works?

-link to docker api doc for this flag
-summarize the change
-what use-cases will this affect?
-do we expect this change have a negative impact on existing implementations?  Please consider for Win/OS X/Linux 

Thanks!

andyrichardson:
Hey @garris, sorry for not checking this earlier.

The justification for using host networking was so that the ports mapped to the host would be accessible to the container at IP `127.0.0.1` (currently, trying to access a host port from the container in Linux is not possible). 

After double checking [the documentation](https://docs.docker.com/network/host/), it looks as though host networking is not supported on macOS/Windows. Whether the addition of this argument will break functionality outside of Linux environments is something I am unsure about. I'll have to experiment more on this but for the time being, I'll close this request.

andyrichardson:
Hey @garris, sorry for not checking this earlier.

The justification for using host networking was so that the ports mapped to the host would be accessible to the container at IP `127.0.0.1` (currently, trying to access a host port from the container in Linux is not possible). 

After double checking [the documentation](https://docs.docker.com/network/host/), it looks as though host networking is not supported on macOS/Windows. Whether the addition of this argument will break functionality outside of Linux environments is something I am unsure about. I'll have to experiment more on this but for the time being, I'll close this request.

andyrichardson:
Hey @garris, sorry for not checking this earlier.

The justification for using host networking was so that the ports mapped to the host would be accessible to the container at IP `127.0.0.1` (currently, trying to access a host port from the container in Linux is not possible). 

After double checking [the documentation](https://docs.docker.com/network/host/), it looks as though host networking is not supported on macOS/Windows. Whether the addition of this argument will break functionality outside of Linux environments is something I am unsure about. I'll have to experiment more on this but for the time being, I'll close this request.

PR #923, opened 2018-11-16
Add KeyPress Interactions to BackstopJS

marcdacz:
I've implemented the KeyPress Interaction for both Puppeteer and Chromy Engines.
I've also added test and updated documentation to include the new command.


marcdacz:
@garris I've updated the documentation. Please have a look. Cheers!

Issue #879, opened 2018-10-02
Time to remove phantom?


garris:
I think it's time to bump to version 4 and remove phantom support.  Users who use Phantom can still use version 3.

Is this going to blow anybody up?  How are we with this?

garris:
And what about Slimer support?   Comments please.  Thanks!

brendonbarreto:
I was think exactly the same thing today. I think we have a few reasons to do that:

- Repository is archived
- The last stable release is from almost 3 years ago

Doing this, maybe we can start thinking about remove `casperjs` as well (CasperJS is no longer actively maintained).

brendonbarreto:
I'm going to have some free time on Friday. Can I start taking a look at this?

garris:
Yes please.   This would be v4.0

Minor note: We'd have to delete references to v3 in docs and ui.

PR #928, opened 2018-11-19
Removing Casper and Chromy

brendonbarreto:
WIP

closes https://github.com/garris/BackstopJS/issues/879


brendonbarreto:
@garris Will we continue with this configuration of engine_scripts? For me it makes sense to remove, because after this PR we will only have the Puppeteer available.

brendonbarreto:
Only now that I realized that on issue you did not mention chromy. Are we going to stay with chromy support?

brendonbarreto:
Creating a more e

garris:
We will keep engine scripts because we will likely add new engines at some point -- e.g. maybe selenium driver.

PR #933, opened 2018-11-27
WIP: load local storage data

Phelilix:
```It provides a place for code to be ran after the first page.goto
local storage data can now be set before the delay and the wait for readySelector.
This commit is mainly so I can get feedback.```

localStorage is only accessible after navigating to the page first.
After setting the authentication data in localStorage using the onReady script (And using page.waitfor) I was able to successfully capture the page I wanted to capture.

The delay and readySelector property is rendered useless if your application needs to use of localStorage. 

It seems clunky so here is my pull request.

What do you think?
I have yet to run this code at this point in time.

### DIFF
--- a/capture/config.default.json
+++ b/capture/config.default.json
@@ -14,12 +14,14 @@
   ],
   "onBeforeScript": "puppet/onBefore.js",
   "onReadyScript": "puppet/onReady.js",
+  "localStorageData": "Does not have precedence over the values in scenarios"
   "scenarios": [
     {
       "label": "BackstopJS Homepage",
       "cookiePath": "backstop_data/engine_scripts/cookies.json",
       "url": "https://garris.github.io/BackstopJS/",
       "referenceUrl": "",
+      "localStorageData": "backstop_data/engine_scripts/localStorage.json"
       "readyEvent": "",
       "readySelector": "",
       "delay": 0,
--- a/capture/engine_scripts/puppet/loadLocalStorage.js
+++ b/capture/engine_scripts/puppet/loadLocalStorage.js
@@ -1,0 +1,40 @@
+var fs = require('fs');
+  
+/*
+  This code is called after the command to navigate has been fired. 
+  before the delay, before the wait for an element to render (readySelector) and before the code that runs just before taking the picture (readyEvent).
+
+  If the data is like this:
+  +------------------------+--------------+
+  |         Key:           |     value:   |
+  +------------------------+--------------+
+  |amplitude_lastEventId   |             2|
+  |amplitude_lastEventTime | 1543336128855|
+  |amplitude_sessionId     | 1543336128827|
+  |amplitude_unsent        |            []|
+  +------------------------+--------------+
+
+  The json should be like this:
+  {
+    "amplitude_lastEventId": "2",
+    "amplitude_lastEventTime": "1543336128855",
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: capture/config.default.json [hunk 1]
```diff
@@ -14,12 +14,14 @@
...
   "onBeforeScript": "puppet/onBefore.js",
   "onReadyScript": "puppet/onReady.js",
+  "localStorageData": "Does not have precedence over the values in scenarios"
   "scenarios": [
     {
       "label": "BackstopJS Homepage",
       "cookiePath": "backstop_data/engine_scripts/cookies.json",
       "url": "https://garris.github.io/BackstopJS/",
       "referenceUrl": "",
+      "localStorageData": "backstop_data/engine_scripts/localStorage.json"
```
garris: Please model this after the load cookies feature.   Please note the cookie param name `cookiePath`.   This should be `localStoragePath` (or something to that effect)

capture/engine_scripts/puppet/loadLocalStorage.js
```diff
@@ -0,0 +1,40 @@
...
+  The json should be like this:
+  {
+    "amplitude_lastEventId": "2",
+    "amplitude_lastEventTime": "1543336128855",
+    "amplitude_sessionId": "1543336128827",
+    "amplitude_unsent": "[]"
+  }
+*/
+
+async function loadLocalStorageData(path, page){
```
garris: Please model this after the `capture/engine_scripts/puppet/loadCookies.js` file.   Guard against bad file paths etc.   

Please implement as an add in -- like `loadCookies` -- see: `capture/engine_scripts/puppet/onBefore.js`

Do not implement this on the main codepath.

core/util/runPuppet.js
```diff
@@ -118,6 +118,19 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
     }
     await page.goto(translateUrl(url));
 
+		//  --- LOCAL STORAGE SCRIPT ---
```
garris: Please implement as an add in -- like `loadCookies` -- see: `capture/engine_scripts/puppet/onBefore.js`

Do not implement this on the main codepath.

capture/config.default.json
Phelilix: I renamed scenarios[0].localStorageData to scenarios[0].localStoragePath, I removed config.localStorageData

Phelilix:
I am still working on it. Currently figuring out how to run the smoke-test in docker.
If I run the test outside docker, I get 68 mismatches, maybe 38 that were exactly the same.

przemyslawjanpietrzak:
Is this PR still open? I need LS feature in my app and I can help with finish requested changes

Phelilix:
Sure, go ahead. 
Do you want me to commit the changes I made after the review?
I theoretically got everything working. But could not test it.
And then butchered it whilst trying to configure docker for it.

I have an inkling of how the dockerfile should be. But I am not sure.

Phelilix:
The dockerfile should be the most simple one...
Where the entire sourcecode is mounted, or the other way around.

przemyslawjanpietrzak:
Yes, I'll be thankful for that. Please ping we when you push commits to your repo, I'll fork it and finish the feature :)

PR #944, opened 2018-12-12
Adding support for a gotoOptions config option

mnemolyst:
Adding support for a gotoOptions config option, to be passed to Chromy, primarily for setting waitLoadEvent:false

### DIFF
--- a/core/util/runChromy.js
+++ b/core/util/runChromy.js
@@ -192,7 +192,12 @@ function processScenarioView (scenario, variantOrScenarioLabelSafe, scenarioLabe
   if (isReference && scenario.referenceUrl) {
     url = scenario.referenceUrl;
   }
-  chromy.goto(url);
+  var gotoOptions = scenario.gotoOptions || config.gotoOptions;
+  if (gotoOptions) {
+    chromy.goto(url, gotoOptions);
+  } else {
+    chromy.goto(url);
+  }
 
   injectBackstopTools(chromy);
 


garris:
Thanks for the PR.  What's the use case for this one?

mnemolyst:
Hi -
Some websites hang on loading and never fire their "load" event, so chromy.goto() blocks until the script times out. Setting {"waitLoadEvent": false} in goto() prevents it from blocking.
Now that I think about it, given the multiple options Backstop provides for delaying screenshots, it might be okay to set waitLoadEvent: false by default.

garris:
That sounds like a pretty big deal -- the load event not firing etc...  

Are you not in a position to stabilize that?

garris:
Does this issue happen using puppeteer engine?

mnemolyst:
It appears so. With Puppeteer I get the following:

> Puppeteer encountered an error while running scenario "..."
> TimeoutError: Navigation Timeout Exceeded: 30000ms exceeded

It's running headless so I can't see just what it's doing, but I believe it's the same error.
If I point it to another url, eventually I see in the console "Browser Console Log 0: JSHandle:BackstopTools have been installed.", which is generated by runPuppet.js on the line after "await page.goto()"
But running against the page that never emits the "load" event, I never see that line in the console.

mnemolyst:
It does succeed with puppeteer if I add { waitUntil: 'networkidle2' } to the page.goto() call

Issue #530, opened 2017-08-31
Help wanted:  add `--filter` arg to backstop approve command


garris:
Would love to be able to approve specific files.

E.g.  backstop approve --filter=<regexStr>   (Regex tests filename)

anton-kulagin:
@garris Hi, i do this stuff but can`t create a branch. So here you can see my commit https://github.com/garris/BackstopJS/compare/master...anton-kulagin:approve-filter?diff=split&expand=1&name=approve-filter

garris:
See PR #534

garris:
closed by #534    Thank you!

PR #945, opened 2018-12-13
docs(readme): update approve filter docs to file name

BBlackwo:
The `backstop approve --filter` needs to be the file name, not the scenario name, as far as I can tell. So just updating the docs appropriately.

Related to #530


BBlackwo:
@garris any update on this?

garris:
@BBlackwo  thank you for the PR!   You are correct, this is actually true.   The only comment I have here is that I think it would help others significantly if we told them where they can find the filenames, which (by default, include scenario and viewport names and) are displayed in the visual and cli reports.

BBlackwo:
Good suggestion @garris. I made that change. Also my editor removed all the trailing spaces 🤷‍♂️.

PR #955, opened 2019-01-14
Making the scrubber sticky

jwongnz:
Some files in compare/output/ got changed automatically.

copy-report-bundle script modified to make the html_report directory before trying to copy to it, as that was causing errors.

Also updating some of the build syntax (; -> &&) https://corgibytes.com/blog/2017/04/18/npm-tips/#bailing-out-when-a-command-fails


BBlackwo:
@garris sure thing. Below is a screenshot and also a [recording](http://recordit.co/LqEgIbJ0OQ):

screen shot 2019-01-15 at 8 50 49 am

garris:
This is a classy update.  Thanks guys!   Will publish as soon as I can.

garris:
Oh, one thing more request.  Sorry.   Looks like you have an older version of `compare/output/diverged.js`.   I think if you rebase the issue should go away.   

Thanks!

BBlackwo:
Hey @garris we tried rebasing, and still have the same issue. I believe the issue is [the script](https://github.com/garris/BackstopJS/blob/master/package.json#L19) that copies the diverged script from the npm module:

```json
"build-compare": "cp ./node_modules/diverged/src/diverged.js ./compare/output/ ..."
```

Might have something to do with [commit 6e6b839](https://github.com/garris/BackstopJS/commit/6e6b8390e547ac9c28c68f366ffa01a52aff0e22#diff-32607347f8126e6534ebc7ebaec4853d) and/or [commit d8a930a](https://github.com/garris/BackstopJS/commit/d8a930a4422f5746f22918605bb8dcc62a41e53d#diff-6f8b68138ad3508525bc772fa054f4f2) where you updated the diverged dependency and the `output/diverged.js` file.

So as far as I can tell, there's a difference between `./node_modules/diverged/src/diverged.js` and what was checked in under `./compare/output/` -- and we've just done a fresh `npm install`, so it seems the issue is the version of `diverged` in the `package-lock.json`. 

So we can revert our changed files, but the next time someone does an `npm install` and then runs `build-compare` the change will come back.

garris:
@BBlackwo ah,  you are correct!   My bad.  I'll deal with it.   You are cleared for takeoff. 😉

PR #960, opened 2019-01-17
Fixing issues with dependencies versioning

brendonbarreto:
I was updating our version of React, when suddenly the images stopped appearing. I undo my changes but still the problem persisted.

And stopping to better analyze the versions that were in `package.lock` and` package.json` I noticed that some were different. 
This can happen because we use `^` in front of the versions. I think it's a good practice to leave our dependencies fixed, to avoid problems like this.

| dependencies             |  devDependencies |
:-------------------------:|:-------------------------:
  |  

In the image above, the dependencies that are marked in the background are the ones that were different. If for some reason I delete my `package.lock` the puppeteer that will be installed will be` 1.11` for example.

@garris Another thing I noticed is that probably the `diverged` version of `npm` is outdated, that's why some PRs are coming with the wrong `diverged.js` file.

It would be nice if you could publish a new version for `npm`, which I will already update in this PR.

Thanks!


garris:
Thank you!

garris:
Hi @brendonbarreto  -- Just fyi -- I had to revert this PR.  Somehow -- this config broke the scrubber 😞.

It's probably something easy to fix -- I just ran out of time to track it down.  Feel free to put up an update.   

Thanks for all your help -- I finally published your version without PhantomJS today 😀.

Cheers.

Issue #979, opened 2019-02-02
puppeteerOffscreenCaptureFix is an undocumented config


pixilz:
**Request**
`puppeteerOffscreenCaptureFix` fixed an issue I was having, but there is no documentation for the config and I am unsure if I should be using it or finding a different solution to the issue I am having.

**Context:**
I recently had an issue where my page had a `margin-top` for a fixed header and some buttons were in the content. I wanted to test the buttons within the container to see hover affects, for some reason the hover was being moved down equidistant to the `margin-top` of my page (causing a button further down than what I wanted to be highlighted). This only occurred on the smallest screen size that I support `320x480`. Setting `puppeteerOffscreenCaptureFix` to `true` fixed this issue. Note that this only happened when `selectors` was set to something that was within the element with the `margin-top`, setting it to `padding-top` did not fix anything.

garris:
Hi @pixilz.   Thank you for posting!    It would be great if you could document this feature in the troubleshooting section of the readme.

PR #980, opened 2019-02-04
Documenting "puppeteerOffscreenCaptureFix"

pixilz:
Added some "troubleshooting" documentation for "puppeteerOffscreenCaptureFix" to make this config more visible.

Addresses #979


patriciamarchand:
Hello,
I would like to try this to fix a problem i have. But i dont understand where use the "puppeteerOffscreenCaptureFix"

PR #984, opened 2019-02-12
Include troubleshooting section about rounding errors

jonathanherdt:
I would have found it very beneficial if the readme would have included something about those weird rounding errors that sometimes pop up so I wrote a short section about that. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -954,6 +954,10 @@ Sometimes bad permissions happen to good people. It's ok, this is a safe space. 
 
 Be sure to use a config `id` in your config file. See https://github.com/garris/BackstopJS/issues/291
 
+### There's no visible difference between "Reference" and "Test" screenshots
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -954,6 +954,10 @@ Sometimes bad permissions happen to good people. It's ok, this is a safe space.
 
 Be sure to use a config `id` in your config file. See https://github.com/garris/BackstopJS/issues/291
 
+### There's no visible difference between "Reference" and "Test" screenshots
+
+You might run into an issue where your reference and test screenshots look seemingly identical but on further inspection have a 1px difference in height or width. To fix this, make sure your viewport is the same as the one you have setup in your configuration. Then, check whether the element you target has a height or width that is a comma-separated value (i.e. 804.25px). If so, you might have run into a rounding error. Adjusting your viewport so the element's width is a natural number will help with this.
```
garris: Hi,  I am not sure this is clear -- the example you give is not comma separated.  Where would a user check this value?  

> height or width that is a comma-separated value (i.e. 804.25px)

jonathanherdt: Hey, I guess I was too much in my German mindset where we use commas instead of dots. I updated the description. Do you think it's useful now?

Also, sorry for taking so long on this, I was working on a different project in the meantime. Next answer won't take so long :)

PR #986, opened 2019-02-14
fix: remove caret from puppeteer version

### LINKED ISSUE #983: Issues with dependency versioning
I'm trying to use a specific version of chrome with backstop, but am unable to do so because of the `^` in front of puppeteer. I see @brendonbarreto made a fix for this in #960 but then it was reverted. Are there plans to include this fix in the future? I think it would be a good idea to bump backstop every time puppeteer is bumped.

aw-davidson:
Minor version changes in puppeteer result in version changes for chrome. Downloading the latest minor version of puppeteer is problematic for projects that skip chromium download.


garris:
@aw-davidson sorry,  I had to revert this.  The report on your PR was broken.   Also,  the new puppeteer has a different rendering behavior -- because of this some of the tests were running inconsistently.

If you'd like to give this another try would you please make sure you can get consistent results when running this 👉 https://github.com/garris/BackstopJS/blob/master/test/configs/backstop_features.js

Thanks!

Issue #987, opened 2019-02-14
Can't pass a config object to the backstop command with Docker


engelfrost:
I have a Backstop project set up where the config is exported as a function which takes an options parameter ([as described in the README](https://github.com/garris/BackstopJS#parse-a-config-file-explicitly)).

I decided to start running tests with Docker, but when I did, running the tests failed. 

```
Delegating command to Docker... docker run --rm -it --mount type=bind,source="/Users/josef/Code/backstopSanityTest",target=/src backstopjs/backstopjs:3.8.8 test "--moby=true" "--config=[object Object]" ""
```

`--config=[object Object]` is of course not a valid configuration. 

Is this an issue with the code or with the documentation?

garris:
I suspect this would work if you were to run backstop locally -- without the docker flag.

I have a feeling that the docker implementation (in backstop) is not compatible with this approach.  My guess is that this would be pretty easy to fix tho.

garris:
Does this work if you were to run backstop locally -- without the docker flag?

I have a feeling that the docker implementation (in backstop) is not compatible with this approach.  My guess is that this fixable if you have the time to look into it.

garris:
Does this work if you were to run backstop locally -- without the docker flag?

I have a feeling that the docker implementation (in backstop) is not compatible with this approach.  My guess is that this fixable if you have the time to look into it.

engelfrost:
Locally it works just fine, it's only when I add the Docker flag it fails. It's the combination of config object literal + Docker that causes the issue.

garris:
Yup,  thanks for confirming that.  I guess to enable this the right way we would want to pass a fully rendered config to the docker instance -- which would take a bit of work.

engelfrost:
Yes, would it be acceptable to generate a temporary config file and delete it once the test is complete?

garris:
Sounds good to me.  Maybe create a temp file with common prefix and unique suffix name in parallel with working config file.  Something like that?

PR #988, opened 2019-02-15
Config object support with Docker

engelfrost:
Solves [garris#987](https://github.com/garris/BackstopJS/issues/987).

It's not possible to pass an object literal via command line to Docker. A config is passed like this: `--config=path-to-config.json`. So the Docker runner assumes that we're using a config file. Currently an object literal evaluated to `--config=[object Object]`.

This is fixed by saving the object as a JSON file (`backstop.config-for-docker.json`) and letting the `config` option point to that file. When the Docker process `exit`s the file is removed, _unless_ the `debug: true` option is passed. Not sure if the debugging thing is actually a good idea.

Does the Docker process always send an `exit` event, even when there is an `error`? If not, the temporary file should also be removed when there is an error.


brownbl1:
We're running into this issue as well. Any chance this can be merged soon?

brownbl1:
We've tested this locally and in our CI environment and it's working well for us, so I definitely vote on merging! Thanks @engelfrost!

garris:
BTW:  Thanks for this PR!  Will work on getting this in after version 4 releases and any issues are sorted out.

webnovationNET:
Do you have ETA for this fix?

garris:
Merged to master! This still needs to be validated -- so it would be great if someone could kick the tires and add a mocha test for it.   

Thanks everyone for your patience -- many apologies for the delay.  

You are awesome.

PR #994, opened 2019-02-22
Ability to read config settings from onBefore script.

jprado:


garris:
Would you please add to documentation?

jprado:
Updated.

jprado:
Good call. Updated.

PR #1007, opened 2019-03-11
openReport config checked when running 'test' via docker

kevinmpowell:
This simple logic change alters the docker branch of `backstop test` so it performs the same config check that `report.js` performs before calling `_openReport`.

As-is, when run via docker, the report is _always_ opened, no matter how the config is set.


adamchenwei:
@garris  @kevinmpowell I thought openReport is not supported in docker instance version of BackstopJS. is the doc out dated in this case? https://github.com/garris/BackstopJS/tree/master/docker

PR #1010, opened 2019-03-15
Include emoji font in docker image 🤓

kevineye:
The official backstopjs docker image does not include any emoji font, so emoji chars show as square boxes in the screenshots.

This change to the Dockerfile installs the "Noto" font, also used by Android for emoji.


tkrah:
@kevineye done with https://github.com/garris/BackstopJS/pull/1470 - can be withdrawn here.

Issue #820, opened 2018-07-12
Take screenshot without resizing page


JamieMcNaught:
Is there any way to take a screenshot without resizing the page? Unfortunately the template I use has some sections which use things like:
.landing-page .page-header {
    height: 100vh;
    position: relative;
}
which unfortunately means that the page-header element becomes huge when taking the screenshot and doesn't represent what it actually looks like on the device.

See the attached screenshots - the shorter two are taken with "Capture Screenshot", scroll, "Capture Screenshot" etc. The longer one is taken with "Capture full size screenshot" - which is I think what BackstepJS is doing. I would expect the longer one to look like the two shorted ones stitched together, but as you can see it's quite different.

Is there any way to get BackstepJS to screenshot without resizing the page (or whatever it is that it's doing).

Hope this makes sense? Many thanks.
![demos creative-tim com_now-ui-kit_index html galaxy s5](https://user-images.githubusercontent.com/1828283/42628830-33ad21ce-85c9-11e8-8372-e885a5b07f1b.png)
![demos creative-tim com_now-ui-kit_index html galaxy s5 2](https://user-images.githubusercontent.com/1828283/42628841-3bee2fae-85c9-11e8-8a4d-f445cc72216b.png)
![demos creative-tim com_now-ui-kit_index html galaxy s5 1](https://user-images.githubusercontent.com/1828283/42628843-3ef98356-85c9-11e8-82f3-1c5cc4c1a139.png)

Jamiewarb:
I'm also experiencing this problem - a fix or workaround would be great!

JamieMcNaught:
Unfortunately using the beta descrived in #785 doesn't seem to be fixing this.

JamieMcNaught:
The problem is really due to the way Chrome tries to take fullscreen screenshots by resizing the browser window.

Here's some code that works around the issue by taking multiple screenshots and stitching together. It also can give you borders around each image which I think may be helpful for determining if the site is fitting correctly to the device screensizes (one of the big issues we'd been having which made us look at some sort of auto testing).

```
const puppeteer = require('puppeteer');
const merge = require('merge-img');

(async () => {
    const vpw = 1024;
    const vph = 768;
    const browser = await puppeteer.launch();
    const page = await browser.newPage();

    await page.goto('https://demos.creative-tim.com/now-ui-kit/index.html');
    await page.setViewport({width: vpw, height: vph});

    const $ele = await page.$('body');
[...]

PR #1029, opened 2019-04-13
workaround incomplete screenshots for large pages by introducing a new (opt-in) config option

sballesteros:
See https://github.com/GoogleChrome/puppeteer/issues/477 
This was inspired by the discussion in https://github.com/garris/BackstopJS/issues/820


juliquiron:
I applied this patch in my work environment and I confirm that it is working fine. 

Thanks @sballesteros for your work here!

garris:
Thanks for this @sballesteros and for checking @juliquiron.   I do not have the bandwidth to merge this now -- but will queue this up.   As long as the new dependency does not raise any security issues this will eventually be merged.  Best, GS

An-dy1:
> As long as the new dependency does not raise any security issues this will eventually be merged. Best, GS

Sweet! I've also been running into this problem.

An-dy1:
It would be awesome to see this fix merged

Its-Alex:
@garris Up ?

garris:
This was manually applied here... https://github.com/garris/BackstopJS/commit/273a9ad74f5b76edaff0944fe8f07216bcab0172

Please note change with `mergeImgHack` flag -- changed to binary switch.  Seems we could fall back to viewport height setting for this?  

```
add mergeImgHack by @sballesteros 

added config option `config.mergeImgHack `

Setting this to true uses merge-img lib to stitch multiple images together instead of taking one large image via puppeteer.

The screenshot height is set at the top of runPuppet.js with `const MERGE_IMG_SEGMENT_HEIGHT = 2000;`

I would prefer this value to be set by the current viewport height set in the config.   (If I am understaning this feature correctly.)

This fixes issue experienced with very large page heights.

Needed to merge this manually because of rebase issue.
```

Install with `npm install backstopjs@beta`

Please validate and provide feedback -- should this become the primary full screen path?

smokris:
(The `mergeImgHack` option was removed in 1c549682c7c9d73bc6e2cb0bde26f96988c5a0da.)

PR #1039, opened 2019-04-29
Fix removeSelectors bug

6uliver:
I added !important to display none to remove elements really. It is necessary because the selected elements can have styles with important too, for example in Bootstrap the d-* helpers:

`<div class="d-flex">I want to be a flex element</div>`

```
.d-flex {
    display: -webkit-box!important;
    display: -webkit-flex!important;
    display: -moz-box!important;
    display: -ms-flexbox!important;
    display: flex!important;
}
```


garris:
Looks good.  Thanks!

PR #1079, opened 2019-08-13
Amendments to text details panel

AkA84:
## Overview
I thought of proposing some UI amendments to the text details panel, based on my experience with working with the tool and while I was at it, I did some refactoring as well (see "Technical details")

### Show label by default instead of the file name

I'd suggest to show the label of the scenario by default instead of the filename. I never found the latter really useful, while the former can be used for:

1. Identify a scenario at a glance (instead of having to mentally extract its label by parsing its filename)
2. Do a quick search for a particular scenario by using its label

In my experience both things really make a difference when you have hundreds of scenarios

Additionally, I added the viewport as an individual field, displayed by default as well alongside the label. Before, the viewport had to be parsed from the filename as well

#### Before
default-before

#### After
default-after

### Show the url also when "text info" is turned on
The "url" field was visible when the user hovered over the filename, but was not visible with the "text info" turned on, making the UI inconsistent (unless there is a reason for it?). 

I've also moved the "selector" field in its own row to improve readibility

#### Before
fields-before

#### After
fields-after

### Improve alignments of test card header elements
More of a side effect of some refactoring I did (see below), now the row with the default info (before the filename, now the label + selector) is horizontally aligned with the nav buttons

#### Before
align-before

#### After
align-after

## Technical details
While working on `<TextDetails />` I noticed that the details were duplicated in the component and it wasn't really clear to me at the beginning why, then I realized it was done in order to cater for both when the details panel is "static" (when then "text info" option is on) and for then it behaves as a popup

```jsx
<WrapperDetails>
  {/* to show in the static version  */}
  <Row hidden={!settings.textInfo}>
    <Label>label: </Label>
    <Value>{label}</Value>
    <Label>selector: </Label>
    <Value>{selector}</Value>
  </Row>
  {/* end */}
  {/* Always visible row, used to open the popup */}
  <Row>
    <Label>filename: </Label>
    <Value onMouseOver={this.showPanel}>{fileName}</Value>
  </Row>
  {/* end */}
  {/* to show in the static version */}
  <DiffDetails suppress={!settings.textInfo} diff={diff} />
  {/* end */}

  {/* popup version, with the same data as above, duplicated */}
  <DetailsPanel {...{ showPanel }} onMouseLeave={this.hidePanel}>
    <Row>
      <Label>label: </Label>
      <Value>{label} </Value>
      <Label>selector: </Label>
      <Value>{selector} </Value>
    </Row>
    <Row>
      <Label>filename: </Label>
      <Value>{fileName} </Value>
    </Row>
    <Row>
      <UrlDetails url={url} referenceUrl={referenceUrl} />
      <DiffDetails diff={diff} />
    </Row>
  </DetailsPanel>
  {/* end */}
</WrapperDetails>
```

So what I did was trying to refactor that bit into a single component that is able to be both "static" and a popup, so that there is no duplication
```jsx
<WrapperDetails {...{ textInfo, showPanel, hidePanel }}>
  <Row>
    <Label>filename: </Label>
    <Value>{fileName}</Value>
  </Row>
  {/* RowDefault is the one always visible */}
  <RowDefault onMouseOver={this.showPanel}>
    <Label>label: </Label>
    <Value>{label}</Value>
    <Label>viewport: </Label>
    <Value>{viewportLabel}</Value>
  </RowDefault>
  <Row>
    <Label>selector: </Label>
    <Value>{selector}</Value>
  </Row>
  <Row>
    <UrlDetails url={url} referenceUrl={referenceUrl} />
    <DiffDetails diff={diff} />
  </Row>
</WrapperDetails>
```
`<WrapperDetails />` is a stateless component that switches between `<WrapperDetailsPopup />` and `<WrapperDetailsStatic />` (both simple styled components) when necessary 
```jsx
const WrapperDetails = props => {
  return showPanel && !textInfo
    ? <WrapperDetailsPopup onMouseLeave={hidePanel}>{children}</WrapperDetailsPopup>
    : <WrapperDetailsStatic textInfo={textInfo}>{children}</WrapperDetailsStatic>;
};
```
It was my first stab at working with styled components, I hope it's good enough! 

### Add a `<CardHeader />` component
In `<TestCard />`, I added a `<CardHeader />` component that wraps both `<TextDetails />` and `<NavButtons />`.

```jsx
const CardHeader = styled.div`
  display: flex;
`;

export default class TestCard extends React.Component {
  render () {
    return (
      <CardWrapper>
        <CardHeader>
          <TextDetails info={info} />
          {!onlyText && (
            <NavButtons />
          )}
        </CardHeader>
        <TestImages />
        <ErrorMessages />
      </CardWrapper>
    );
  }
}
```

This was necessary to maintain the vertical spacing between the header and the screenshots whenever the newly refactored `<TextDetails />` switches to popup mode (= it gets `position: absolute;`)

#### Without `<CardHeader />`
before

#### With `<CardHeader />`
after

As a side effect, it improved the alignment of the header elements as showed above

### Use `border` property instead of `:after` pseudoselector
Not sure why that was the case, but the green/red border on a test card was `:after` pseudoselector, while it can simply be an actual border of the card?

Couldn't see any major issue after making the change, but maybe I missed something

#### Before
```jsx
const CardWrapper = styled.div`
  &:before {
    content: '';
    display: block;
    width: 8px;
    height: 100%;
    background-color: ${props => props.status === 'pass' ? colors.green : colors.red};
    position: absolute;
    top: 0;
    left: 0;
  }
`;
```

#### After
```jsx
const CardWrapper = styled.div`
  border-left: 8px solid transparent;
  border-left-color: ${props => props.status === 'pass' ? colors.green : colors.red};
`;
```


garris:
Wow.  This is great -- Thank you for the updates & cleanup!

AkA84:
@garris how come this was reverted?

garris:
There were a few display issues.  

In particular
- when filenames were very long the Pop-up ran off the screen.  
- in some conditions hovering over the pop up caused the display to shift
- there is a usecase where we would not want viewport to display
- I think we may want very robust truncation in most fields in most situations-- but we would also want the ability to see that data with a second interaction so that it can be copied either partially or in whole

Sorry for the delay here.  Hope that makes sense.

If it is helpful I can provide information on how to reproduce some of these cases.

AkA84:
@garris yes please, if you could provide repro steps for all of those cases it would be very helpful

AkA84:
@garris 

> when filenames were very long the Pop-up ran off the screen.

I can't seem to be able to replicate it, the filename seems to be wrapping as expected in the popup, both before and after my PR

**Before**
before

**After**
after

> in some conditions hovering over the pop up caused the display to shift

I'd need to do which conditions and what you mean by "shift"

> there is a usecase where we would not want viewport to display

I'd need to know which usecase

> I think we may want very robust truncation [...]

Honestly I'm not sure I understood what you're talking about here, nor if it's something related to my changes or not

Thanks!

PR #1080, opened 2019-08-13
Add approve to list of scripts

AkA84:
While working on https://github.com/garris/BackstopJS/pull/1079 I noticed that `approve` is the only backstop command that is not exposed in "scripts".

I thought that maybe it's an oversight, so I'm adding it


AkA84:
@garris is this of any use to you?

garris:
Yes.  Thanks for adding this. 👍

Issue #1093, opened 2019-09-06
Can we add a baseUrl for the HtmlReport?


svwu:
For several integrations it would be awesome to have the option to add an own basurl 
which are used for the html report.

The Urls generated in the config.js always add "../" behind all paths, which does not fit all CI Integrations which support an html report.

I would love to have an option to manipulate the "../" to a path on my own choice.

PR #1094, opened 2019-09-06
Added option to override the report base path for resources

svwu:
For our CI Integration, we need to specify the baseUrl where the Report loads resources from.
It may be an edge case but it's really necessary for us.

I opened an Issue aswell:
https://github.com/garris/BackstopJS/issues/1093


svwu:
We fixed the problem in our CI with some workarounds - this PR is not necessary for us anymore.
If it is useful for others, it can be considered to be still merged or not.

garris:
Thanks @svwu.   Just curious how you solved the issue.  Please post if you have a second.  Cheers!

PR #1103, opened 2019-10-19
Copy viewport objects

gbasto:
Consider the following configuration:
```javascript
const defaultViewports = [
  {
    label: 'sm',
    ...
  },
  {
    label: 'md',
    ...
  },
  {
    label: 'xl',
    ...
  },
];

const backstopConfig = {
   viewports: defaultViewports,
   scenarios: [
     {
        label: 'scenario A',
        ...
     },
     {
        label: 'scenario B (only runs on xl viewport)',
        viewports: defaultViewports.filter(viewport => viewport.label === 'xl'),
        ...
     },
     {
        label: 'scenario C',
        ...
     },
   ],
   ...
};
```
Given that the viewport index is saved in the viewport object itself, in the previous example, both 'sm' and 'xl' will always have the index 0, because the `scenario B` viewports config will override the global 'xl' index.

With this PR, it's created a copy of each viewport configuration, making possible to reuse viewports configuration with predictable viewports indexes.

### DIFF
--- a/core/util/createBitmaps.js
+++ b/core/util/createBitmaps.js
@@ -82,7 +82,7 @@ function decorateConfigForCapture (config, isReference) {
 }
 
 function saveViewportIndexes (viewport, index) {
-  return Object.assign({}, viewport, {vIndex: index});
+  return Object.assign({}, viewport, { vIndex: index });
 }
 
 function delegateScenarios (config) {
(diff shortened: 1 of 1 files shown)


garris:
Looks good.  Did you run tests?

garris:
Did you add a test for your use-case?

PR #1127, opened 2019-12-04
Added option cliExitOnEngineError to bubble up if engine has an error…

mummybot:
**Exposing engine errors**
By default if there is an error in the engine e.g. missing selector, it won't exit BackstopJS. This is normally fine as the screenshots will pick up any unexpected differences. If you want to terminate BackstopJS on engine errors set the option flag:

```json
"cliExitOnEngineError": true
```


garris:
Hi there @mummybot -- apologies for the delay.     

I believe backstop is correctly reporting errors in cases where something in a scenario goes wrong.  

Can you help me understand your situation better so we can figure out the right approach?

garris:
Ok.  Got it.  I understand your thinking here -- but may I ask that we not short-circuit the bitmap generation here.  The errors that occur during this flow should still propagate up.  Would it be possible for you to handle this the same way that `test` handles it?  (Only without actually comparing files).   You can see how this is done in `core/util/compare/index.js` -- The code is unfortunately really undocumented -- could use some inline comments. 

Please let me know if that would be possible.  Thanks!

garris:
Sounds good.  This would be an improvement which I don't expect would break existing implementations -- so I am thinking default functionality.

core/command/index.js
```diff
@@ -20,7 +20,8 @@ var commandNames = [
   'test',
   'approve',
   'version',
-  'stop'
+  'stop',
+  'engineErrors'
```
garris: I think you should be able to just add this as a regular module (from `/core/util`).  We should avoid creating this as a command since this would never be run alone.

core/command/reference.js
```diff
@@ -5,6 +5,7 @@ const { shouldRunDocker, runDocker } = require('../util/runDocker');
 
 module.exports = {
   execute: function (config) {
+    const executeCommand = require('./index');
```
garris: not needed

core/command/reference.js
```diff
@@ -22,6 +23,7 @@ module.exports = {
         return createBitmaps(config, true);
       }).then(function () {
         console.log('\nRun `$ backstop test` to generate diff report.\n');
+        return executeCommand('_engineErrors', config);
```
garris: you should be able to just run a plain old module. like here...   https://github.com/garris/BackstopJS/blob/master/core/command/test.js#L10

core/command/engineErrors.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  execute: function (config) {
```
garris: You can just export the anon function here.  Skip the execute property and the object all together.

core/command/index.js
mummybot: Done

core/command/reference.js
mummybot: Yup, agreed

core/command/reference.js
mummybot: Done

core/command/engineErrors.js
mummybot: Done

garris:
published to backstopjs@4.4.0

`npm install backstopjs@canary`

Please validate when you get a chance -- thanks!

PR #1143, opened 2020-01-26
Bumps pupeteer to v2

orta:
👋 - I'd like to test the dark mode of the TypeScript website https://github.com/microsoft/TypeScript-Website/pull/187 - to do that I need an API which was added in puppeteer 2.0

It's [puppeteer v2](https://github.com/puppeteer/puppeteer/releases) because it drops node 6 support:

> Puppeteer now requires Node.js v8+; Node.js v6 is no longer supported

I couldn't find a note in the docs about what versions you support for Node, so it's really your call on whether this is a big enough change to warrant a backstop major version. Doesn't seem to be an API changes but maybe the Chrome upgrade could introduce rendering diffs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5850,9 +5850,9 @@
       "dev": true
     },
     "https-proxy-agent": {
-      "version": "2.2.4",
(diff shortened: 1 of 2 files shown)


garris:
Next BackstopJS version will drop node versions below 8.

PR #1147, opened 2020-02-07
Format apt-get commands according to best practices

t3easy:
See: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

Resolves: 1004


t3easy:
t3easy/backstop                         latest                  0582e29b35a3        3 minutes ago       1.88GB
backstopjs/backstopjs                latest                  e4cc7b3e2802        11 days ago         1.9GB

It's not that much but a first start.

PR #1150, opened 2020-02-20
Feature/approve button

1276stella:
- Add the functionality in the browser report to approve a test
- Fix eslint errors in one file
- Update the readme doc

**Approve a test**
approve-btn-demo

**Display an error message when approving a test fails**
Screen Shot 2020-02-18 at 10 02 36 PM


compare/src/components/molecules/ApproveButton.js
```diff
@@ -0,0 +1,116 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import styled from 'styled-components';
+import { approveTest, filterTests } from '../../actions';
+import { colors, fonts } from '../../styles';
+
+const REMOTE_HOST = 'http://127.0.0.1';
+const REMOTE_PORT = 3000;
```
shankarsridhar: There is an [open issue](https://github.com/garris/ember-backstop/issues/34) in `ember-backstop` to make backstop run on a port other than the default port 3000. Will having this as a constant make it harder to implement that ?

compare/src/components/molecules/ApproveButton.js
```diff
@@ -0,0 +1,116 @@
...
+
+  approve () {
+    const { fileName } = this.props;
+    const url = `${REMOTE_HOST}:${REMOTE_PORT}/approve?filter=${fileName}`;
+    this.setState({ approveStatus: 'pending' });
+
+    fetch(url, {
+      method: 'POST'
+    })
+      .then(response => {
```
shankarsridhar: Can we use `async`  `await` here for better readability ?

compare/src/components/molecules/ApproveButton.js
```diff
@@ -0,0 +1,116 @@
...
+        if (response.ok) {
+          this.setState({ approveStatus: 'approved' });
```
shankarsridhar: Can this string literal be imported from a common mapping or constants location `approved`? instead of hard coding it in multiple places(there is at least one more occurence down this file) ?

core/util/remote.js
```diff
@@ -0,0 +1,61 @@
...
+ * @param {String} originalJsonpReport - jsonpReport `report(${jsonReport});`
+ * @param {String} approvedFileName - the name of the screenshot that is approved
+ * @return {String} jsonpReport - modified jsonpReport
+ */
+function modifyJsonpReportHelper (originalJsonpReport, approvedFileName) {
+  const report = extractReport(originalJsonpReport);
+  report.tests.forEach(test => {
+    if (test.pair.fileName === approvedFileName) {
+      test.status = 'pass';
+      delete test.pair.diffImage;
```
shankarsridhar: Can `test.pair` be undefined or falsy here ?

core/util/remote.js
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Modify the test status of the JSONP report based on the approved file name. JSONP is used
+ * to create the Backstop report in browser. This function ensures the UI consistency after
+ * a user apporves a test in browser and refreshes the report without running a test.
+ *
+ * @param {Object} params - the input params
+ * @param {String} params.reportConfigFilename - the path to the html report config file
+ * @param {String} params.approvedFileName - the name of the screenshot that is approved
+ * @return {Prmoise}
```
shankarsridhar: Typo

core/util/remote.js
```diff
@@ -0,0 +1,61 @@
...
+    })
+    .catch(err => {
+      throw new Error(`Failed to modify the report. ${err.message}.`);
+    });
+}
+
+module.exports = {
+  modifyJsonpReport: modifyJsonpReport,
+  modifyJsonpReportHelper: modifyJsonpReportHelper
+};
```
shankarsridhar: Why not use the shorthand syntax ?
https://alligator.io/js/object-property-shorthand-es6/

```js
module.exports = { modifyJsonpReport, modifyJsonpReportHelper };
```

remote/index.js
```diff
@@ -1,114 +1,149 @@
...
+    app._backstop.tests[req.params.testId] = {
+      [req.params.scenarioId]: req.body
+    };
+
+    console.log(
+      'Processing dynamic test request for ',
+      `dview/${req.params.testId}/${req.params.scenarioId} `,
+      app._backstop.testCtr
+    );
+    console.log('Loadinfg dynamic config template at ' + PATH_TO_CONFIG);
```
shankarsridhar: Typo:
`Loadinfg ` => `Loading`

compare/src/components/molecules/ApproveButton.js
1276stella: Hi Shankar, I am using a map 'APPROVE_STATUS_TO_LABEL_MAP'. 'approved' here is the key of the map.

shankarsridhar: 👍 got it. it did not look obvious in the first glance.

compare/src/components/molecules/ApproveButton.js
garris: This is a great point.   When Backstop opens a report (either internally or via user on cli) it runs openReport.js -- notice here is where the port is defined...
https://github.com/garris/BackstopJS/blob/724f16e76ae353b7b37cc18915c053d03b73c2af/core/command/openReport.js#L9
So, it is probably just a matter of pulling out the dynamic port property out of the config at that point and passing into the above port definition.

AN ADDITIONAL THING TO NOTE:  This also happens to be where we explicitly tell the report UI that it is in REMOTE mode.   Some features, like `Approve` will only work in remote mode (i.e. when backstopRemote is running).  I don't remember if the UI has a global variable to track this -- but it needs one.  And the `Approve` option should not be shown unless we are in remote mode.

garris: @1276stella ☝️

core/util/remote.js
garris: wow -- what a catch.

compare/src/components/molecules/ApproveButton.js
1276stella: I see. I moved the object to the top of the file and used uppercase for the keys. Hope this could help.

core/util/remote.js
1276stella: Sure

core/util/remote.js
1276stella: `tests` has the following format. This function is triggered when a user clicks on the `Approve` button, which means there is a valid test. So I don't see any scenarios that could have no `test.pair`.

```
  "tests": [
    {
      "pair": {
        "reference": "../bitmaps_reference/backstop_default_BackstopJS_Homepage_0_document_0_phone.png",
        "test": "../bitmaps_test/20200219-232153/backstop_default_BackstopJS_Homepage_0_document_0_phone.png",
        "fileName": "backstop_default_BackstopJS_Homepage_0_document_0_phone.png",
        ...
      },
      "status": "pass"
    },
   ...
]
```

PR #1153, opened 2020-02-26
doc(usage): --filter works for backstop reference

mick352:


mick352:
You're welcome! Thanks for this awesome tool.

PR #1167, opened 2020-03-26
Fix ugly bug with setViewport function

Patrycja20:
Function 'setViewport' is async. So, we have to write await before call async function.
This bug causes errors with actions on resized pages. Espiecially on smaller screens on phone.


ZielonyBuszmen:
I had the errors with tests in mobile resolution (clicks, fill inputs, etc), and this resolve those problems.

I approve this changes! Thanks :smiley: :+1: :heavy_check_mark:

PR #1172, opened 2020-04-03
feat(remote): use report paths from config

fuhlig:
Instead of using a hardcoded (default) path for the report,
use the path set in the backstop config. This allows custom paths.

### DIFF
--- a/remote/index.js
+++ b/remote/index.js
@@ -108,9 +108,7 @@ module.exports = function (app) {
       });
 
       const reportConfigFilename = path.join(
-        PROJECT_PATH,
-        'backstop_data',
-        'html_report',
+        _config.paths.html_report,
         'config.js'
       );
       await modifyJsonpReport({


fuhlig:
❓
I am not sure what [this function does](https://github.com/garris/BackstopJS/blob/master/remote/index.js#L23) and if it's needs to change as well to take paths set in the config

```js
app.use(function (req, res, next) {
    req.url = req.url
        .replace(/\/backstop\/backstop_data/, '/backstop_data') // change this with path set in config?
        .replace(/\/backstop\/dview/, '/dview')
        .replace(/\/backstop\/dtest/, '/dtest');
    next();
});
```

garris (approved):
Yes,  I think this change should work for ordinary usage with backstop. And hopefully it should work with ember-backstop as well.

As a separate note -- the 'test' here flow should probably be enhanced to take a config object also -- I noticed here that a config is not passed.  Not asking you to make this change necessarily-- just a general note.

To answer your other question-- The URL replacements are there because of a specific use-case in ember-backstop where some URLS will come with `backstop/` prepended.   In that case -- it *is* possible you have found a latent issue where changing paths outside `backstop_data` will break.  This should probably be tested -- but this is not a show stopper for your change.

Apologies for the delayed review -- all last year was one long crisis management exercise for me.  Glad the world seems to be on the path to recovery!

Thanks again!

Issue #1166, opened 2020-03-23
Switching between `passed` and `failed` tabs breaks UI state of the Approved/Approve button


shankarsridhar:
**TL;DR:**
- Display logic for `Approve` button seems to be inconsistent with the backing data for the UI.
- Filtering is not working unless page is refreshed.

After backstop tests are run, and backstop report page is opened (by running `ember backstop-report`), the `Approve` button,  if present, comes back after switching between the `passed` and `failed` tabs.

Just to be clear, here is what I am referring to as the `passed` and `failed` tabs:
![image](https://user-images.githubusercontent.com/4473214/77271852-f7ed1900-6c6c-11ea-8de5-1775ff0c4d46.png)

*Expectation:*
- The `Approve` button should not be present for tests that have already been approved.
- The test that was approved (by clicking on `Approve` button) should be filtered as passed.

*What actually happens:*
- The `Approve` button is present for tests that have already been approved.
- Filtering tests bases on passed/failed seems to be _not_ working. Tests that have been approved are still displayed in the failed category.  

If the backstop report page is refreshed, then all the UI state seems to be updated/rendered correctly as per the backing data.

Here is a screen capture to illustrate the problem:
[Approve-button-bug.mov.zip](https://github.com/garris/BackstopJS/files/4366817/Approve-button-bug.mov.zip)

garris:
Great catch @shankarsridhar!

@1276stella is this because we are not saving `approved` state? (my guess is we simply update the DOM)

1276stella:
@garris a PR is opened to fix this issue. Please see the detailed explanations in the PR. https://github.com/garris/BackstopJS/pull/1175 Thank you!

PR #1175, opened 2020-04-09
fix(report): #1166 approve button does not work on the filtered failed page

1276stella:
Solve issue #1166 
There are two issues:
1. The card slot, instead of the card itself, retains the `approveStatus`. On the filtered failed page, when I approve the 1st card, its approveStatus changes from `Initial` -> `Pending` -> `Approved`. After `filterTests` is dispatched, the 1st card is removed from the page. However, its `Approved` status remains on the page. Fix: change `Approved` status to `Initial`.
2. The index of the card on the current filtered page is used to find the right test to approve. But this is not correct since the same card on the `all` page and the `failed` page has different indices. Fix: use `fileName` to find the test to approve.

Before
Before

After
After


Alexbirvik:
@garris @1276stella Hi guys! I don't have button "approve" in my reports and I didn't find any information how to add it. Your help well be super appreciated.

PR #1176, opened 2020-04-15
feat: add ability to test focus states

quagliero:
This update allows you to test focused elements in the same way as click and hover elements. Useful
for ensuring no regressions for keyboard users and a11y checking around buttons and interactive
controls.


examples/Jenkins/Sample/backstop_data/engine_scripts/puppet/clickAndHoverHelper.js
```diff
@@ -8,6 +9,11 @@ module.exports = async (page, scenario) => {
     await page.hover(hoverSelector);
   }
 
+  if (focusSelector) {
```
garris, line 12: Just curious here -- any reasoning with ordering after hover and before click?

quagliero: No real reason other than a mental model of how I view events (hover and focus more likely to offer small UI updates of intent, and clicks actually doing something). In my experience from using it on our projects, we used hover _or_ focus on our tests, (one simulating mouse, the other a keyboard) so it was more just ensuring it came before the click.

Serhansolo:
Is there any update when this will be merged?

ryuran:
Hello @garris, It could be useful to have this feature 👍 
Nice work @quagliero

Issue #1180, opened 2020-04-26
NPM Audit alerts with chromy


broksonic21:
After adding backstopjs via `npm install --save-dev backstopjs`, two new `npm audit` alerts show up. Is there a way to skip chromy install (we use puppeteer exclusively), or otherwise resolve?

```
┌──────────────────────────────────────────────────────────────────────────────┐
│                                Manual Review                                 │
│            Some vulnerabilities require your attention to resolve            │
│                                                                              │
│         Visit https://go.npm.me/audit-guide for additional guidance          │
└──────────────────────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ minimist                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=0.2.1 <1.0.0 || >=1.2.3                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ backstopjs [dev]                                             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ backstopjs > chromy > chrome-launcher > mkdirp > minimist    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1179                            │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ minimist                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=0.2.1 <1.0.0 || >=1.2.3                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ backstopjs [dev]                                             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ backstopjs > chromy > jimp > mkdirp > minimist               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1179                            │
└───────────────┴──────────────────────────────────────────────────────────────┘```

garris:
[...]

PR #1181, opened 2020-04-27
Remove chromy as per discussion in https://github.com/garris/BackstopJS/issues/1180

broksonic21:
This is my first commit in BackstopJS repo, so apologies in advance if over or under-aggressive on this. Happy to tweak up.

This removes all references to chromy I could find as per discussion in https://github.com/garris/BackstopJS/issues/1180

One test case json came out different in the smoke tests; I think as I changed one test from chromy to puppeteer. Let me know if that should be reverted.

Note: While this addresses the chromy `npm audit` findings, I did not run `npm audit --fix` when installing locally for this PR, but I noticed a number of alerts in the devDependencies around this.


core/util/createBitmaps.js
```diff
@@ -133,17 +127,7 @@ function delegateScenarios (config) {
...
-    var getFreePorts = require('./getFreePorts');
-    return getFreePorts(PORT, scenarioViews.length).then(freeports => {
-      console.log('These ports will be used:', JSON.stringify(freeports));
-      scenarioViews.forEach((scenarioView, i) => {
-        scenarioView.assignedPort = freeports[i];
-      });
-      return pMap(scenarioViews, runChromy, { concurrency: asyncCaptureLimit });
-    });
-  } else if (config.engine.startsWith('puppet')) {
+  if (config.engine.startsWith('puppet')) {
```
garris, line 130: Could we leave the `/chrom./` case and throw and error here -- something like -- _Chromy is no longer supported in version 5+.  Please use version 4.x.x for chromy support._

ahaverbuch: updated!

ahaverbuch:
I ran smoke-tests, sanity-test (similar errors, only file updated looks just due to system difference and is included here) and unit-test (all passed). Integration-tests didn't run for me.

Is there more needed?

core/util/createBitmaps.js
broksonic21: updated!

broksonic21:
I ran smoke-tests, sanity-test (similar errors, only file updated looks just due to system difference and is included here) and unit-test (all passed). Integration-tests didn't run for me.

Is there more needed?

core/util/createBitmaps.js
```diff
@@ -129,6 +129,9 @@ function delegateScenarios (config) {
 
   if (config.engine.startsWith('puppet')) {
     return pMap(scenarioViews, runPuppet, { concurrency: asyncCaptureLimit });
+  }
+  else if (/chrom./i.test(config.engine)) {
+    logger.error(`Engine "${(typeof config.engine === 'string' && config.engine) || 'undefined'}" not recognized! Chromy is no longer supported in version 5+. Please use version 4.x.x for chromy support.`);
```
garris, line 134: Maybe this should just be: "Chromy is no longer supported in version 5+. Please use version 4.x.x for Chromy support."  I don't think you need the first bit with all the type checking and stuff.  We already know inside this block that we have a chromy user who probably bumped their version and now they are confused about why everything is failing.  Please let me know if there is something I am missing here.  Thank you!

broksonic21: updated

broksonic21:
Reverted jsonReport and updated error message. Once this merges, happy to make a follow-on PR to get rid of some of the local `npm audit` issues seen (mostly in dev dependencies) that just need updating. Didn't include here for separation of concern reasons.

garris:
Looks great -- thanks for all the deleting -- I feel so much lighter now!

Issue #1156, opened 2020-03-02
Using the newest Docker BackstopJS image breaks our tests (Jenkins)


jonathanherdt:
Our builds currently break with the newest Docker BackstopJS image, 4.5.0. It breaks with both npm version `4.4.2` and `4.5.0`.

Out of our ~1100 tests, 395 fail. Judging from how the tests fail, it seems like BackstopJS doesn't wait long enough until the content is rendered. Most screenshots are completely white and a few show content which is not completely rendered. Most of the tests that are failing are simple HTML/CSS pages. They load some JavaScript but it does not change the appearance. We also have parts of our pages built with React. With the React components, the "white screenshot" issue happens with a much smaller percentage but it is still there.

The original of the screenshot I'm showing here can be seen here: https://www.static-immobilienscout24.de/fro/core/4.2.3/documentation/html/campaign-box.html#usage

If need be, I could also provide a zip file with the HTML pages and our `backstop.json`.

![Screen Shot 2020-03-02 at 12 31 04](https://user-images.githubusercontent.com/6302908/75672752-b1b22480-5c81-11ea-9536-b671a348cd50.png)

garris:
That doesn't sound good.  This happened when you bumped your version?

garris:
Does the smoke test run for you?  https://github.com/garris/BackstopJS#smoke--feature-tests

jonathanherdt:
We didn't bump it but we had `image: backstopjs/backstopjs` (without a version number) in our `docker-compose.yml`.

jonathanherdt:
I wouldn't know how to run the smoke test in a local installation with a custom `docker-compose.yml`. I tried pointing the script to the correct config file (`node_modules/backstopjs/test/configs/backstop`) but then lots of files are missing.

I also just cloned the BackstopJS repo and ran the smoke tests from there. There, it looked fine.

jonathanherdt:
Also of note: We're using a custom `docker-compose.yml` with the following content:

```yml
version: "2.1"
services:
    testrunner:
        image: backstopjs/backstopjs:4.5.0
        volumes_from:
            - container:${HOSTNAME}:rw
        working_dir: ${WORKSPACE}
        shm_size: 2048m
        command: "test  --config ./src/test/regression/.generated.backstop.json"
```

We're running it with the following command: `docker-compose -f docker-compose.yml run --rm testrunner`

engelfrost:
[...]

PR #1196, opened 2020-05-26
Docker BackstopJS sometimes produces white or cut-off screenshots

kasparsz:
Fix for #1156

### DIFF
--- a/core/util/runPuppet.js
+++ b/core/util/runPuppet.js
@@ -422,13 +422,15 @@ async function captureScreenshot (page, browser, selector, selectorMap, config, 
         const box = await el.boundingBox();
         if (box) {
           // Resize the viewport to screenshot elements outside of the viewport
-          const bodyHandle = await page.$('body');
-          const boundingBox = await bodyHandle.boundingBox();
+          if (config.useBoundingBoxViewportForSelectors === false) {
+            const bodyHandle = await page.$('body');
+            const boundingBox = await bodyHandle.boundingBox();
 
-          await page.setViewport({
-            width: Math.max(viewport.width, Math.ceil(boundingBox.width)),
-            height: Math.max(viewport.height, Math.ceil(boundingBox.height)),
-          });
+            await page.setViewport({
+              width: Math.max(viewport.width, Math.ceil(boundingBox.width)),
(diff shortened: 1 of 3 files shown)


kasparsz:
Docker of course fails since code in the backstopjs/backstopjs is out of my control

kasparsz:
What the fix does is resizes the viewport so that it takes full body size just before taking a screenshot of an element. This is done only for elements since full page screenshots were already working properly

garris:
There is already an offscreen capture fix -- what makes this one better?

kasparsz:
puppeteerOffscreenCaptureFix doesn't work
Added a comment with more details https://github.com/garris/BackstopJS/issues/1156#issuecomment-634964511

core/util/runPuppet.js
```diff
@@ -421,8 +421,18 @@ async function captureScreenshot (page, browser, selector, selectorMap, config,
       if (el) {
         const box = await el.boundingBox();
         if (box) {
+          // Resize the viewport to screenshot elements outside of the viewport
```
garris, line 424: @kasparsz   Thank you for this thoughtful solution.  Would you mind adding a feature flag here? Maybe call it `useBoundingBoxViewportForSelectors` we can default to "on" e.g. 
```
if (config.useBoundingBoxViewportForSelectors === false) {
    Your resize fix...
}
```
This way -- if someone has an issue we can advise them to disable this feature.

garris: You'll need to handle for `undefined` etc.   Please add tests for this also.

Thank you!

kasparsz: Thank You

Commit: https://github.com/kasparsz/BackstopJS/commit/e8fda8ded539a814d0bf730deb0a9fab6d112bce

```if (config.useBoundingBoxViewportForSelectors !== false) { ... fix ... }```
Added, but since it's on by default, anything except for "false" would enable it, so"undefined" would still mean it's on, so "!== false" should be sufficient in that case. Or did I misunderstood you?

Test puppet_backstop_features_hideSelectors_0_document_0_phone.png before this fix was failing already consistently, a test for this would be a copy of it.

garris:
I just tested this manually.  It's great!   I will update the tests after the docker build goes out.

Issue #1210, opened 2020-07-17
Gray Blank and Mostly Blank/Gray images getting captured


jmickela:
It seems that this is similar to the issue that led to the introduction of the puppeteerOffscreenCaptureFix feature, however it's not fixing the captures for me. I've managed to reproduce the issue with the following config. When I run backstop reference and backstop test the result is nothing but gray boxes. 

```json
{
  "id": "backstop_default",
  "viewports": [
    {
      "label": "phone",
      "width": 320,
      "height": 2000
    },
    {
      "label": "tablet",
      "width": 1024,
      "height": 2000
    }
  ],
  "onBeforeScript": "puppet/onBefore.js",
  "onReadyScript": "puppet/onReady.js",
  "scenarios": [
    {
      "label": "Hotmail?",
      "url": "https://outlook.live.com/owa/",
      "referenceUrl": "",
      "readyEvent": "",
      "readySelector": "",
      "delay": 0,
      "hideSelectors": [],
      "removeSelectors": [],
      "hoverSelector": "",
      "clickSelector": "",
      "postInteractionWait": 0,
      "selectors": [
        ".premium-appeal"
      ],
      "selectorExpansion": true,
      "expect": 0,
      "misMatchThreshold" : 0.1,
      "requireSameDimensions": true
    }
  ],
  "paths": {
    "bitmaps_reference": "backstop_data/bitmaps_reference",
    "bitmaps_test": "backstop_data/bitmaps_test",
    "engine_scripts": "backstop_data/engine_scripts",
    "html_report": "backstop_data/html_report",
    "ci_report": "backstop_data/ci_report"
  },
  "report": ["browser"],
  "engine": "puppeteer",
  "puppeteerOffscreenCaptureFix": true,
  "engineOptions": {
    "args": ["--no-sandbox"]
  },
  "asyncCaptureLimit": 1,
  "asyncCompareLimit": 1,
  "debug": false,
  "debugWindow": true
}
```

PR #1211, opened 2020-07-19
fixes more issues with capturing elements out of the viewport

jmickela:
This is a fix for issue #1210. The possible drawback to this is that it expands the height of the viewport. 

It's piggy-packing on the puppeteerOffscreenCaptureFix config, which isn't necessary, but it seemed overkill to add another edge case config item that is addressing the same issue. However, if expanding the height of the viewport isn't acceptable when using puppeteerOffscreenCaptureFix, it would be helpful if we could add another config item to enable this code since I'm getting nothing but gray boxes for things outside the viewport without it.

### DIFF
--- a/core/util/runPuppet.js
+++ b/core/util/runPuppet.js
@@ -435,6 +435,13 @@ async function captureScreenshot (page, browser, selector, selectorMap, config, 
           var type = config.puppeteerOffscreenCaptureFix ? page : el;
           var params = config.puppeteerOffscreenCaptureFix ? { path: path, clip: box } : { path: path };
 
+          if(config.puppeteerOffscreenCaptureFix) {
+            const pageHeight = await page.evaluate("document.body.scrollHeight");
+            await type.setViewport(
+              Object.assign({}, type.viewport(), {height: parseInt(pageHeight) })
+            );
+          }
+
           await type.screenshot(params);
         } else {
           console.log(chalk.yellow(`Element not visible for capturing: ${s}`));


megclaypool:
This works for me, and addresses white boxes/clipping for selectors below the original viewport.

I also adapted this code to address clipping of the full-page, "document" selector (which is handled in another conditional). 

If you add the following to line 406(ish), between the `if (fullPage && config.mergeImgHack) {...}` and the `else {...}`, your "document" screenshots will come out properly too:
```
else if (fullPage && config.puppeteerOffscreenCaptureFix) {

      const pageHeight = await page.evaluate("document.body.scrollHeight");
      await page.setViewport(
        Object.assign({}, page.viewport(), {height: parseInt(pageHeight) })
      );

      try {
        await page.screenshot({
          path: filePath
          // ,fullPage: fullPage
        });
      } catch (e) {
        console.log(chalk.red(`Error capturing..`), e);
        return fs.copy(config.env.backstop + ERROR_SELECTOR_PATH, filePath);
      }
}
```

vegaskev (approved):
Based on my understanding of `browser` and `page` in backstop, this looks like it will solve the vh issue, without causing any other conflicts.  I'm relatively new to the library's codebase, however, I do not see any issues with this.

We've also tested this solution locally, and found success, without any side effects.

garris:
@vegaskev @megclaypool   Does this solution require `config.puppeteerOffscreenCaptureFix` to be true?   If so -- are we able to simply enable `config.puppeteerOffscreenCaptureFix` by default?   Will that cause any regressions on implementations where `config.puppeteerOffscreenCaptureFix` was never enabled?

PR #1216, opened 2020-07-30
Using interactive mode and reference mode in backstop remote

shankarsridhar:
Backstop already supports config to use interactive mode but it was not passed on to backstop remote.

**Changes include:**
- Ability to switch between reference and test modes in backstop remote
- Enable option to use interactive mode in backstop remote

### DIFF
--- a/remote/index.js
+++ b/remote/index.js
@@ -3,7 +3,8 @@
 
 var parseArgs = require('minimist');
 var argsOptions = parseArgs(process.argv.slice(2), {
-  string: ['config']
+  string: ['config'],
+  boolean: ['i']
 });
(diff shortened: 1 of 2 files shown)


remote/index.js
```diff
@@ -66,7 +66,8 @@ module.exports = function (app) {
       vid: app._backstop.testCtr
     };
 
-    backstop('test', { config }).then(
+    const command = req.path.includes('dtest') ? 'test' : 'reference';
```
garris, line 69: I wonder if we should put leading and trailing slashes around `/dtest/` to like, make sure data in `:testId` and `:scenarioId` don't accidentally trip the test?

PR #1218, opened 2020-08-04
Improve handling edge cases

shankarsridhar:
This PR addresses some comments(from https://github.com/garris/BackstopJS/pull/1216) around handling some edge cases in a better way.

**Changes:**
* better handling of boolean flag - incremental flag
* stricter check for `/dtest` based on which reference or test modes are selected.

_P.S: Tested against _ember-backstop-tutorial_ repo._

### DIFF
--- a/remote/index.js
+++ b/remote/index.js
@@ -14,6 +14,8 @@ var express = require('express');
 var backstop = require('../core/runner');
 var { modifyJsonpReport } = require('../core/util/remote');
 
+const booleanizeArg = incrementalFlag => [true, 'true'].includes(incrementalFlag);
+
 module.exports = function (app) {
   app._backstop = app._backstop || {};
   app._backstop.testCtr = 0;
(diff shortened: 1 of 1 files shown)


remote/index.js
```diff
@@ -14,6 +14,8 @@ var express = require('express');
 var backstop = require('../core/runner');
 var { modifyJsonpReport } = require('../core/util/remote');
 
+const booleanizeArg = incrementalFlag => [true, 'true'].includes(incrementalFlag);
```
garris, line 17: What happens when incrementalFlag is `"false"`?

remote/index.js
```diff
@@ -66,8 +68,8 @@ module.exports = function (app) {
       vid: app._backstop.testCtr
     };
 
-    const command = req.path.includes('dtest') ? 'test' : 'reference';
-    backstop(command, { config, i: Boolean(req.body.i) }).then(
+    const command = req.path.includes(`/dtest/`) ? 'test' : 'reference';
```
garris, line 71: Would you mind instead checking for `/dref/` so that default behavior is `test`?  I know it's not currently a concern, but we may alias more routes to this flow.

remote/index.js
shankarsridhar: This method(`booleanizeArg()`) will return boolean `false` when the input  is anything other than `true` or `'true'`.

remote/index.js
shankarsridhar: I've updated the PR accordingly.

remote/index.js
garris: awesome.

PR #1222, opened 2020-09-03
Update Chrome cookie extension link in Readme

dpihlaja:
Existing link was 404, linkrot strikes again

Replaced with an extension that mass exports in same .json format


garris:
Thank you very much for this PR.  I updated the docs in a separate PR.   Best, Garris

Issue #1225, opened 2020-09-07
Failure due to missing library libXss.so.1 in Docker image


merelrietbergen:
Hi!

In the latest (5.0.4) Docker image (build September 5, 2020) the library `libXss.so.1` is not installed, resulting in a failure with the following error:
```
COMMAND | Error: Failed to launch the browser process!
                /usr/local/lib/node_modules/backstopjs/node_modules/puppeteer/.local-chromium/linux-722234/chrome-linux/chrome: error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory
```

Adding `libxss1` to the libraries being installed in the `Dockerfile` seems to resolve the issue, though there may be some other underlying cause.

Neither 5.0.1 nor 5.0.2 are affected, likely because the images for these were not built as recently.

digitaldonkey:
Ran into same issue.

jlnschfr:
same here, occured today.

`/usr/local/lib/node_modules/backstopjs/node_modules/puppeteer/.local-chromium/linux-722234/chrome-linux/chrome: error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory`

AdamChrimes:
Having the same issue on Travis CI

```
COMMAND | Executing core for "test"
createBitmaps | Selected 89 of 89 scenarios.
      COMMAND | Command "test" ended with an error after [0.033s]
      COMMAND | Error: Failed to launch the browser process!
                /usr/local/lib/node_modules/backstopjs/node_modules/puppeteer/.local-chromium/linux-722234/chrome-linux/chrome: error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory
                
                
                TROUBLESHOOTING: https://github.com/puppeteer/puppeteer/blob/master/docs/troubleshooting.md
                
                    at onClose (/usr/local/lib/node_modules/backstopjs/node_modules/puppeteer/lib/Launcher.js:750:14)
                    at Interface.helper.addEventListener (/usr/local/lib/node_modules/backstopjs/node_modules/puppeteer/lib/Launcher.js:739:50)
                    at Interface.emit (events.js:194:15)
                    at Interface.close (readline.js:379:8)
                    at Socket.onend (readline.js:157:10)
                    at Socket.emit (events.js:194:15)
                    at endReadableNT (_stream_readable.js:1125:12)
                    at process._tickCallback (internal/process/next_tick.js:63:19)
[...]

PR #1226, opened 2020-09-09
Clean up dockerfile.

digitaldonkey:
I think you need to combine your installs or it will try to install without souces/lists.


digitaldonkey:
More heavy idea from https://medium.com/@cloverinks/how-to-fix-puppetteer-error-ibx11-xcb-so-1-on-ubuntu-152c336368

digitaldonkey:
fix #1225

garris:
BOOM FIXED!

PR #1235, opened 2020-09-23
fix Topbar import in Header component

### LINKED ISSUE #1234: Incorrect import path of Topbar in Header.js
As mentioned in #1154 in the Header component the import of Topbar component is incorrectly using a lowercase path `import Topbar from '../organisms/topbar';`.
Case-sensitive filesystems are running into compile errors with this import.
```
ERROR in ./compare/src/components/ecosystems/Header.js
Module not found: Error: Can't resolve '../organisms/topbar' in '/home/developer/BackstopJS/compare/src/components/ecosystems'
 @ ./compare/src/components/ecosystems/Header.js 21:14-44
 @ ./compare/src/components/App.js
 @ ./compare/src/index.js
```

The import should be changed to the correct path `import Topbar from '../organisms/Topbar';`

I also noticed that `index_bundle.js` is compiled in development mode, is that intended or was it mistakenly checked in that way?

jKatt:
This PR fixes #1234

The huge div in `index_bundle.js` is mainly because it is compiled in production mode now.


garris:
Thanks for this!

Ok to try a prod build option.  Just please check that it works 😉.

jKatt:
thanks.

Can you please release a new version with this fix?
The prod build is kind of useful when integrating html reports in jenkins as it is not using eval like the dev build, which is much more CSP friendly.

garris:
@jKatt quick question, how did you generate the prod build? I would assume we would want to add a flag to our local build config?  I didn't see that in this commit.

jKatt:
@garris it is the default since it is not specified in the `webpack.config.js`.
For consistency `mode: 'production'` could be added to it.

jKatt:
I made a PR #1236 to add it to the webpack config.

PR #1236, opened 2020-09-25
add mode to webpack config

jKatt:
This PR will add `mode: 'production'` to the webpack config.
Since it is the default nothing should break.

### DIFF
--- a/compare/webpack.config.js
+++ b/compare/webpack.config.js
@@ -1,6 +1,7 @@
 const path = require('path');
 
 module.exports = {
+  mode: 'production',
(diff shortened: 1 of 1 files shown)


ledenis:
I approve this PR 👍. This removes the below warning when running `npm run build-and-copy-report-bundle`.

> WARNING in configuration
> The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
> You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/configuration/mode/

PR #1252, opened 2020-10-31
feat: set default userAgent for viewports

bratva:
Set UserAgent by default is a good practice, because some server can detected that and response special bundle for client.

### DIFF
--- a/capture/engine_scripts/puppet/onBefore.js
+++ b/capture/engine_scripts/puppet/onBefore.js
@@ -1,3 +1,4 @@
 module.exports = async (page, scenario, vp) => {
   await require('./loadCookies')(page, scenario);
+  await require('./setUserAgent')(page, scenario, vp);
 };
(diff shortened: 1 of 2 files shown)


fuhlig:
- instead of relying on hardcoded values like `phone`, consider making it configurable
- instead of relying of "custom" and hardcoded values for userAgents, puppeteer has better options by using [`puppeteer.devices`](https://pptr.dev/#?product=Puppeteer&version=v5.5.0&show=api-puppeteerdevices)

Snippet from docs
```js
const puppeteer = require('puppeteer');
const iPhone = puppeteer.devices['iPhone 6'];

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.emulate(iPhone);
  await page.goto('https://www.google.com');
  // other actions...
  await browser.close();
})();
```

Issue #1254, opened 2020-11-06
Allow additional YML config file


digitaldonkey:
Maintaining a distribution the backstop.json file is limited as JSON does not allow comments. 

To explain to my users (and develop more easily) it would be great if I could provide a backstop.yml file as alternative.

garris:
You can easily use a .JS file instead. You can use comments in there.  check the docs for more info.

digitaldonkey:
Thank you for clarification.

PR #1255, opened 2020-11-09
Improve documentation for config file

digitaldonkey:
I didn't see the JS config file option. Would consider it misplaced. Added an example.  
I would prefere a automatically picked up YML file, but technically this closes my question in  #1254


garris (approved):
Awesome -- thank you for improving the ReadMe Doc!

PR #1258, opened 2020-11-16
Remove cookie domain as it could be munged from scenario url

digitaldonkey:
When trying to make our our pipeline multi domain compatible we stumbled into the cookie domain setting in provided by default in cookies.json. With a little fix in loadCookies.json you can avoid specifying domain per cookie. 

I think it might be a good default setting.

### DIFF
--- a/capture/engine_scripts/cookies.json
+++ b/capture/engine_scripts/cookies.json
@@ -1,6 +1,5 @@
 [
   {
-    "domain": ".www.yourdomain.com",
     "path": "/",
     "name": "yourCookieName",
     "value": "yourCookieValue",
--- a/capture/engine_scripts/puppet/loadCookies.js
+++ b/capture/engine_scripts/puppet/loadCookies.js
@@ -1,7 +1,7 @@
 const fs = require('fs');
 
 module.exports = async (page, scenario) => {
-  let cookies = [];
+  const cookies = [];
   const cookiePath = scenario.cookiePath;
 
   // READ COOKIES FROM FILE IF EXISTS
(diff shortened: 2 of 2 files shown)


garris (approved):
Looks good!   Thanks.

capture/engine_scripts/cookies.json
```diff
@@ -1,6 +1,5 @@
 [
   {
-    "domain": ".www.yourdomain.com",
```
garris, line 3: Only problem I have here is that this property will be there when using popular plugins to create this JSON file.   So, I prefer to leave this as I imagine many users will include this.

capture/engine_scripts/puppet/loadCookies.js
```diff
@@ -1,22 +1,22 @@
...
 
-  // MUNGE COOKIE DOMAIN
+  // Inject cookie domain, url, secure from scenario url.
+  const url = new URL(scenario.url);
   cookies = cookies.map(cookie => {
-    if (cookie.domain.startsWith('http://') || cookie.domain.startsWith('https://')) {
-      cookie.url = cookie.domain;
-    } else {
-      cookie.url = 'https://' + cookie.domain;
+    cookie.url = `${url.protocol}//${url.hostname}`;
```
garris, line 15: I prefer this way more to what is currently here.  I am a little conflicted about using the "magic" hostname since this could differ from the one exported in the cookie-- but I think this is your whole point.  So, I am supportive of this because it will probably work more often than not.

garris (changes requested):
Just one thought re: `domain` property in the cookies.json file.

PR #1262, opened 2020-12-05
fix scrubber isolated ref image

### LINKED ISSUE #902: big reference image in report points to wrong folder
hi,

being on the report page, reference image preview shows correctly, while on click, it opens the test image

this was pointed out in https://github.com/garris/BackstopJS/issues/850#issuecomment-432704234 but I think it should have its own issue ID

ledenis:
Fix #902

## To reproduce the issue:

Edit `test/configs/backstop_fail_cases.js`

```diff
     {
       label: 'expect',
       url: '../../index.html',
       selectors: ['p'],
       selectorExpansion: true,
       expect: 8,
+      misMatchThreshold: 100
     }
```

(so we can see different images on tests that "pass")

Then:

```
cd test/configs/
node ../../cli/index.js test --config=backstop_fail_cases --filter=expect
```

In the report, on a test that "passed", click on the "reference" image 
## Actual behavior
The "test" image is opened

Demo of actual behavior

## Expected behavior
The "reference" image should be opened instead

This is fixed with this PR


garris (approved):
Looks good to me!

garris:
Thank you for cleaning this section up!

PR #1268, opened 2021-01-07
Fix issue with mergeImgHack resulting in whitespace

### LINKED ISSUE #1267: `mergeImgHack: true` results in white space
With long pages on the latest (5.0.7) when `mergeImgHack: true` only the first capture contains any of the webpage, the rest is all white space. This happens both whether running via docker or not for me, with a fresh docker image install.

This is the netflix homepage where the only deviation from the initial default config is to enable `mergeImgHack`.

netflixcom_desktop-old

DEfusion:
Enabling `mergeImgHack` was resulting in whitespace being produced for all but the first screenshot. To resolve this I discovered that the page needed to be scrolled to the appropriate part of the page for the screen.

It's not entirely clear to me why this is needed, I did a lot of exploring in the debug browser window to figure out if there was a better solution but I couldn't find anything although I'm sure there must be.

For comparison here is a before & after capture of netflix's homepage:

| Master      | This branch |
| ----------- | ----------- |
| netflixcom_desktop-old      | netflixcom_desktop       |

Fixes #1267

### DIFF
--- a/core/util/runPuppet.js
+++ b/core/util/runPuppet.js
@@ -1,4 +1,3 @@
-const MERGE_IMG_SEGMENT_HEIGHT = 2000;
 const puppeteer = require('puppeteer');
 
 const mergeImg = require('merge-img');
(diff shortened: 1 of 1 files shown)


garris (approved):

garris:
This is great! Thank you very much for investigating this and providing a fix!

PR #1273, opened 2021-01-22
Remove slang from readme.md

conradkirschner:


garris (commented):
Thank you for this.   I updated the language in another commit.   I will close this request.  Best.  Garris

PR #1294, opened 2021-03-30
Update Dockerfile node version to 12.19.0

genepaul:
Node 10 is end-of-life as of April, so the Dockerfile should at least be on a Node version that is still in maintenance mode.


genepaul:
I have built and tested this dockerfile locally to verify that it works. I'm not sure if there are any tests I need to update for this, or what else is needed to get this merged in.

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10.15.3
+FROM node:12.19.0
```
rmkanda, line 1: Latest is 12.22.0

garris:
Thank you for this.  @rmkanda  mentioned the the latest is two minor bumps ahead -- but this is fine with me.

PR #1316, opened 2021-05-11
refactor: replaced resolve dependency with require.resolve()

### LINKED ISSUE #1315: ReDoS in unmaintained path-parse sub-dependency of dependency resolve
As even already [reported to the `resolve` dependency maintainers](https://github.com/browserify/resolve/issues/240), there's a [CVE opened for the](https://nvd.nist.gov/vuln/detail/CVE-2021-23343) sub-dependency [path-parse](https://www.npmjs.com/package/path-parse), which even already gets reported in tools like WhiteSource.
Quick summary of that ticket: It’s been closed as the [path-parse](https://www.npmjs.com/package/path-parse) maintainers have been alerted on that issue, but as the package seems to be unmaintained this might not get resolved (soon).

The [`resolve`](https://www.npmjs.com/package/resolve) dependency has been introduced with commit 9b9cb161bd4c68f73083c3d20c083fdc4f70af4b to "find SSWS during remote cmd":
https://github.com/garris/BackstopJS/blob/e5b3d49e7869a96d7e1e9f6b33778421c39845d5/core/command/remote.js#L5

My question is: What for do we need this dependency, instead of using `require.resolve()` directly? Probably I’m missing something here, at least the tests didn’t show any related problem, and the path string that is used for the upfollowing `node` command was the same as the one generated via `resolve`.

mfranzke:
Closes #1315

Replaced `resolve` depency with `require.resolve()`

### DIFF
--- a/core/command/remote.js
+++ b/core/command/remote.js
@@ -1,8 +1,7 @@
 const logger = require('../util/logger')('remote');
 const path = require('path');
 const { exec } = require('child_process');
-const resolver = require('resolve');
-const ssws = resolver.sync('super-simple-web-server', { basedir: __dirname });
+const ssws = require.resolve('super-simple-web-server');
 
 module.exports = {
   execute: function (config) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,22 +5,28 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0",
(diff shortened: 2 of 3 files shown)


garris:
Thanks for this update. Backstop remote is a dependency for ember-backstop.  I am hoping this will not hose anything.

Cheers!

Issue #528, opened 2017-08-31
Small diff not picked up when mismatchThreshold is 0


luujam1:
For a very small change in a large area with mismatchThreshold set to 0, the diff is not detected:

![screen shot 2017-08-31 at 12 03 19](https://user-images.githubusercontent.com/8289865/29920088-d8f57362-8e43-11e7-85de-894049618329.png)

However when a bit more of the component is changed the diff is detected:

![screen shot 2017-08-31 at 12 04 20](https://user-images.githubusercontent.com/8289865/29920156-1680e27a-8e44-11e7-8154-4323db2aec65.png)

I'd expect any difference to be detected when mismatchThreshold is set to 0. Any help on this would be much appreciated. Thanks.

garris:
I would expect this too.  Inthink the workaround would be to set threshold to 0.001

garris:
There is probably a guard in there intended to set the threshold to a default level if threshold is undefined.  But it's probably doing that in the case of zero also.   Would love a PR if you can resolve it.

joseph-stano:
I think this is similar to https://github.com/garris/BackstopJS/issues/477.

@luujam1 - would you agree?

luujam1:
@joseph-stano yeah that looks like the same issue.

PR #1337, opened 2021-07-08
use resemble's more accurate rawMisMatchPercentage value

### LINKED ISSUE #477: misMatchThreshold of 0 still allows changes on large reference images
Nice work on Backstop!

We are diff'ing a project with a lot of data on a particular page. The reference image is quite large (1024x23046).

Even with a misMatchThreshold=0, we've noticed very small changes to the image will not fail the test. A single offset word, for example. 

We've tried other misMatchThreshold permutations like 0.00, 0.000000001, etc. and the tests still pass.

I'm guessing this difference is too small of a percentage for resemble.js, fractions of a percent. In which case this is not a Backstop issue, and likely resemble.js. But I thought I'd start here in case we're missing something.

PhilippMi:
instead of the rounded misMatchPercentage string value as suggested in #477

fixes #477 and #528


garris:
Will this break existing implementations?

PhilippMi:
This change makes backstop more sensitive, so tests which previously passed could fail afterwards (if `misMatchThreshold` is below 0.01).

garris:
@PhilippMi -- If this is the case, I think we need to put this behind a config flag so we don't get a thousand engineers posting about BackstopJS behaving unexpectedly.

Would you be willing to add a flag to the config?  I guess it would be something like `useRawMisMatchValue`?

PhilippMi:
Good idea! I'll change it.

PhilippMi:
@garris Do you still need something?

garris:
Sorry - I didn't see this update.  Looks good!

Issue #1208, opened 2020-07-13
Backstop hangs on last scenario when using "readyEvent"


cvolpe13:
We've got a static Gatsby site running and we're using backstop for visual regression. For the most part it has been working pretty well for us. 

One thing I wanted to improve is the hard coded delays that are in place via use of the `readyEvent` and so we implemented a console log accordingly. The problem that I have and cannot figure out is backstop appears to hang indefinitely when using `readyEvent`. At first I thought it would hang in random spots but then realized it was also hanging on the very last scenario as defined in the `backstop.json` file. 

Things I've tried:
* Running with and without docker 
* Running with an async capture and compare limit of 1 but this didn't help (it would still hang)
* Running a single scenario works fine with `readyEvent` 

Has anyone else experienced this, looking for some insight. 

Thanks

garris:
One common issue here is if your console event fires prior to backstop initializing in your app.  This is probably what is happening.  Potentially, instead of using the console you can use `readySelector`.  If you need an arbitrary event you can put a noop class on your body that backstop would wait for.

cvolpe-ecobee:
Thanks @garris. So it sounds like I should forgo using  `readyEvent`. I'll try with `readySelector` and see how it goes. We're not too keen on the idea of a noop class, but going to see if this may work for us.

Thank you

DEfusion:
I'm having the same issue with backstop latest, even if I try delaying it (or outputting it multiple times with a recursive call and `setTimeout`) it never gets also it never causes BackstopJS to timeout.

I tried switching to ready selector which also isn't working for me, though at least it does time out `TimeoutError: waiting for selector ".-modal-ready" failed: timeout 30000ms exceeded` again no matter how early or late I ensure the selector is there.

masi:
I had some issues with readyEvent and started to look for workarounds. One solution was to hack a timeout into BackstopJS so that a failing test would block the suite which is a PITA with automated testing.
The other approach was to try a custom readyProperty within onBefore. The idea is to wait for a global JS-Variable being set by the page's code.

raDiesle:
Here i have the same issue that after last scenario will finish when using any readyEvent will not stop process.

cvolpe13:
> Here i have the same issue that after last scenario will finish when using any readyEvent will not stop process.

Ya, `readyEvent` is still a no go for me. I'm still stuck on using a hard coded wait. @garris Anymore thoughts/ideas on this?

masi:
[...]

PR #1338, opened 2021-07-08
Add readyEventTimeout functionality. readyEvent hangs sometimes.

klodoma:
Related to #1208

When readyEvent is used an not fired, BackStopJs Hangs. 

Added a readyTimeout for readyEvent or readySelector properties. Default 30000ms. 
If the readyEvent or readySelector is not fired within this time the promises will be resolved.


garris (changes requested):
@klodoma would you be open to renaming this parameter to `readyTimeout` such that this value will be used for either `readyEvent` or `readySelector`?

klodoma:
Sure, that's a good idea. I'll adjust it.

klodoma:
Ok, I am using just readyTimeout now. 

`await page.waitForSelector(scenario.readySelector);` has a default timeout of 30000ms so I was not sure if I should put the `readyTimeout` as a parameter to it. Could be an option. 

Please review.

garris:
Yes.  Please add `readyTimeout` as a parameter to `waitForSelector`.

Also,  can you please create a test for this?  Maybe set the timeout to 5 seconds and test that scenarios behave as expected.

klodoma:
Please check the latest changes. I've added the tests too. 

Is it normal that the reference files are not up to date? Should I update/commit them?
I am getting quite some differences:

Screenshot_77

core/util/runPuppet.js
```diff
@@ -72,10 +72,16 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
...
   const readyEvent = scenario.readyEvent || config.readyEvent;
-  let readyResolve, readyPromise;
+  const readyTimeout = scenario.readyTimeout || config.readyTimeout || 30000;
+  let readyResolve, readyPromise, readyTimeoutTimer;
   if (readyEvent) {
     readyPromise = new Promise(resolve => {
       readyResolve = resolve;
+      // fire the ready event after the readyTimeout
+      readyTimeoutTimer = setTimeout(() => {
+        console.error(chalk.red('*** readyEvent Timeout! ***'), scenario.url);
```
garris, line 82: Could we make this log more explicit?  
_ReadyEvent not detected within readyTimeout limit. (\`${readyTimeout}\` seconds)_

garris:
Also,  when you ran the tests -- did you use the docker version of the tests?  I think those text differences will go away if you use the docker version.

eg `npm run smoke-test-docker `

core/util/runPuppet.js
klodoma: Fixed.

klodoma:
> eg `npm run smoke-test-docker `

It hangs if I leave the new tests in, but that makes sense. If I remove the new tests, it run through, there are still 2 diffs. See screenshot.

Screenshot_79

garris:
@klodoma Yes, if those are the only two differences it's totally fine.  Thank you!

klodoma:
> @klodoma Yes, if those are the only two differences it's totally fine. Thank you!

@garris yes, only those 2.

garris (approved):
LGTM

klodoma:
Would it be possible to release this asap?

garris:
Done

```
npm notice === Tarball Details ===
npm notice name:          backstopjs
npm notice version:       5.3.5
npm notice package size:  22.8 MB
npm notice unpacked size: 28.2 MB
npm notice shasum:        dd9c7d29d998c7d8de2214be67121343c11156eb
npm notice integrity:     sha512-+ucYnXjl+UAb/[...]IXj1AuHYwSFVg==
npm notice total files:   527
npm notice
+ backstopjs@5.3.5
```

klodoma:
Thank you!

PR #1350, opened 2021-08-10
Feature/code inspections

klodoma:
Changes:

- run idea64 code inspection
- reformat code with eslint
- convert var to const/let

This is only about code formatting, getting rid of several warnings in the editor. 

I haven't checked all of them, first I would like to get a feedback if this direction is OK or NOT.


conradkirschner:
I would like to see that in the next major version, maybe also with some git hooks implemented to force codestyle. 

For my pullrequest I tried to keep the current code-style, to make it easier to merge/read... 
https://github.com/garris/BackstopJS/pull/1342

garris (approved):
Thank you -- nice cleanup work!

Issue #1349, opened 2021-08-09
Change proposal


klodoma:
In the current situation the `html_report` contains always the latest report and the older reports are lost. 

I propose to add a "flag" where the `html_report` folder gets a similar structure to the `bitmaps_test` folder (with timestamps as sub-folders) which contains then the actual reports. 

This helps if one executes the test repeatedly and want to see the diffs. 

Any thoughts against such a feature?

![Screenshot_6](https://user-images.githubusercontent.com/944350/128719885-ddcbb413-5c39-438e-af44-b8d6171e2af3.png)

garris:
@klodoma yes, I am supportive.  Others have asked for this as well.  It could be as simple as an operation which runs after the current reporting flow completes and copies all dependent assets into an additional html_report directory structure. 

SideNote:  I personally don't have a big problem with redundant assets floating around to ensure backwards compatibility -- so pure copy functions are fine in this solution are with _me_ -- but those concerns sometimes come up for others.  That said, I would approach any concerns of cleaning up old/redundant files as a separate housecleaning problem.

klodoma:
@garris I agree with redundancy, this is not something one will always use. 

We could add a flag like: `keepReports: true` or `reportsHistory: true`, I would go with the former.

garris:
I like `keepReports` -- but maybe `archiveReports` is a little more accurate (cause we would literally be creating self-contained report archives.)

nicolasgasco:
Hello, everyone! Is this feature still available? We're struggling with generating an HTML report that includes all timestamped directories that are generated and not simply the latest one.
We tried activating the flag introduced in the MR attached to this issue, but to no avail.

PR #1353, opened 2021-08-14
Feature/archive report

klodoma:
Archive reports functionality. Based on #1349. 

@garris please review the config variable naming's, I am not 100% happy yet. 

We'll need 2 variables:
 - `archiveReport:true/false`
 - and in "paths"
```
"paths": {
...
    "reports_archive": "backstop_data/reports",
  }
```
The reports are generated in `reports` folder, see screenshot:
Screenshot_8

### DIFF
--- a/core/command/report.js
+++ b/core/command/report.js
@@ -51,7 +51,11 @@ function archiveReport (config) {
 
   return fs.copy(toAbsolute(config.html_report), archivePath).then(function () {
     const file = path.join(archivePath, path.basename(config.compareConfigFileName));
-    return replaceInFile(file, /"..\\\\/g, '"../../');
+    // replace the "..\\" with "..\\..\\" in the config.js files
+    // on windows double escape in order to work properly
+    const search = path.sep.replace(/\\/g, '\\\\\\\\');
+    const replace = path.sep.replace(/\\/g, '\\\\');
+    return replaceInFile(file, new RegExp(`"..${search}`, 'g'), `"..${replace}..${replace}`);
   });
 }
 
--- a/core/command/report.js
+++ b/core/command/report.js
@@ -7,6 +7,23 @@ const fs = require('../util/fs');
 const logger = require('../util/logger')('report');
 const compare = require('../util/compare/');
 
+function replaceInFile (file, search, replace) {
+  return new Promise((resolve, reject) => {
+    fs.readFile(file, 'utf8', function (err, data) {
+      if (err) {
+        reject(err);
+      }
+      const result = data.replace(search, replace);
+
+      fs.writeFile(file, result, 'utf8', function (err) {
+        if (err) reject(err);
+      }).then(() => {
+        resolve();
+      });
+    });
+  });
+}
+
 function writeReport (config, reporter) {
   const promises = [];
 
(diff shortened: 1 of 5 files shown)


core/command/report.js
```diff
@@ -88,7 +125,13 @@ function writeBrowserReport (config, reporter) {
       throw err;
     });
 
-    return allSettled([jsonpConfgWrite, jsonConfgWrite]);
+    let promises = [jsonpConfgWrite, jsonConfgWrite];
+
+    if (config.archiveReport) {
+      promises.push(archiveReport(config));
```
garris, line 131: Hey, are `jsonpConfigWrite` and `jsonConfigWrite` guaranteed to have completed when we call `archiveReport()`?  I _think_ these are dependencies.

garris (changes requested):
Left a comment -- I _think_ those two config generation flows need to resolve before running the copy command.   Please give that a check.  Thanks for working on this feature!

core/command/report.js
klodoma: Good point. Fixed.

conradkirschner:
Just as a addition to this feature: 

How can I open the report and switch between the version of the reports? (Feature request for report frontend?)

garris:
@conradkirschner I assume one could just look at the archive directory and point their browser to different directories -- which i assume will have a full html report as index.html?  With backstopRemote you could probably make a simple index page by crawling everything at the path specified at `config.archivePath`.   Does that make sense?

klodoma:
> @conradkirschner I assume one could just look at the archive directory and point their browser to different directories -- which i assume will have a full html report as index.html? With backstopRemote you could probably make a simple index page by crawling everything at the path specified at `config.archivePath`. Does that make sense?

We could indeed create an `index.html` in `config.archivePath` that we could re-generate every time a new report is available.
Probably there is a package that we could use to generate this on the fly.

conradkirschner:
Thought more about some buttons in the report "Next Report[date]" "Prev Report[Date]" + a dropdown with the history. 

for such things, It would be nice to have a static config that get generated if archive is set to true, that contains some infos about the existing ones

klodoma:
> Thought more about some buttons in the report "Next Report[date]" "Prev Report[Date]" + a dropdown with the history.
> 
> for such things, It would be nice to have a static config that get generated if archive is set to true, that contains some infos about the existing ones

This is a very good idea I think. This will change quite a bit the way the reporting is being done. 
Maybe we should discuss it and do it in a next iteration? I would not push it in this PR.

garris:
I agree that this would be a nice-to-have feature in some other iteration.  That said, I would vote for a feature that would auto-delete archives (and test bitmap directories) after n-days would be a more popular feature.

klodoma:
@garris do we need something else here? I pushed the requested changes some days ago.

garris (approved):

garris (approved):
This looks great!

garris:
Sorry for the delay -- Merged!

> @garris do we need something else here? I pushed the requested changes some days ago.

garris:
@klodoma There is an issue here that I hope you can resolve.   This archiving feature depends on the `bitmaps_reference` and `bitmaps_test` image directories to remain in the same state.  If I enable this feature and subsequently update my reference directory then the reference previews in my archive becomes invalid.   In addition, if I clean out my test directory then I may be missing test images all together.   This is why I think we need to be duplicating test and reference assets locally inside each report directory.   I proposed this earlier but maybe this point did not come across.   Is this something you think you could address?

klodoma:
@garris I see your point. I think this could be seen as another step into the `archive reporting`

My initial target was to have something like now: execute multiple tests (without changing the references) and be able to see the reports. 

I'll have a look how can we also duplicate also the reference files.

PR #1355, opened 2021-08-15
Feature/npm package updates

klodoma:
An initial step into updating the npm packages. I've tried a big-bang approach, but the react app failed to work, so I'll take a step-by-step approach. 

## Changes

- update image reference files
- update eslint version
- fix eslint notifications
- add .gitlab-ci.yml as alternative to travis(ran out of travis credits) to be able to execute CI tests

 
### Side-note:
I have another branch, where I've added `sanity-test` and `smoke-test` as part of testing. This branch is based on the `npm-package-updates' one, so it's easy to merge. I ran the tests in gitlab.

See here: https://github.com/klodoma/BackstopJS/tree/feature/travis-updates
 ```
script:
  - npm run lint && npm run unit-test
  - npm run sanity-test
  - npm run smoke-test
```

Gitlab pipeline: https://gitlab.com/klodoma/BackstopJS/-/pipelines/353523538


garris (approved):
This looks good.   Thank you for cleaning this up.

garris:
@klodoma   I think that `smoke-tests` has some scenarios that sometimes throw false positives (they are not well written).   Before adding to CI we'd need to skip the smoke-tests or fix the smoke tests.   The `sanity-test` is very simple and probably good to go and we can add this -- but it is really only checking very simple behavior.

> I have another branch, where I've added `sanity-test` and `smoke-test` as part of testing.

garris (commented):
LGTM

Issue #1352, opened 2021-08-13
Idea: no-cache meta tag


conradkirschner:
Why? 

I am using github pages to access the report and was wondering why it isn't reloading. 

After I disabled the cache, page got reloaded. 

To avoid such things it might be a idea to disable cache completly? (really not sure) 

``` html
<meta http-equiv="cache-control" content="no-cache"/>
```

@klodoma What do you think?

klodoma:
@conradkirschner yes, it would make sense to use `no-cache` for the report files. 

I think you should use something like this for more browser compatibility:
```
<meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate">
<meta http-equiv="Pragma" content="no-cache">
<meta http-equiv="Expires" content="0">
```

We could make a PR and push it in the master.

PR #1357, opened 2021-08-16
- Disable Cache on client side

conradkirschner:
Reason: 

We set it to no cache in the report, because on github pages we don't have control over the caching strategy. 

Caching here is not needed. 

https://github.com/garris/BackstopJS/issues/1352


garris:
Thank you for this PR.   Cheers.

Issue #1362, opened 2021-09-22
Logging capture


SeanMcMillan:
I'd like to have a way to keep a per-run log file.

I am way off the beaten path; I've got a web app that lets our ops people run Backstop against our staging sites before and after changes are deployed to make sure nothing unexpectedly changes. Unfortunately, it runs multiple sites in parallel, so the raw console logs are one big mishmash of several simultaneous runs. I'd like to be able to provide a logger object when I call `backstop()`. I'd love a structured log where I can see which scenarios gave which output, or maybe even a log report of some kind.

I know this is a rare use case. I plan to make a PR for it myself, but I thought I would drop an issue here first to solicit feedback on what the API for that should look like.

SeanMcMillan:
For my use case, I really want to be able to see the logs for a given test case, (or test case + viewport pair), which makes **push log info into the reports** a better approach for me. However, my servers are flaky (or _something_ is anyway,) so I also want to see logs for the reference action.

Maybe it would be simplest to drop the logs right next to the snapshots, so where I have `test-case_viewport.png`, I'll have `test-case_viewport.log` right next to it.

Then we could turn it on with something like `saveLogsPerScreenshot: true` at either the top level of the config, or in the `scenarios` section. (Final key name TBD.)

garris:
This is a good start.  My comments would be...

1) in the UI, if log data is present in a scenario, add a button to show that scenario's log data.  When the button is clicked, the log data is displayed in an overlay which contains a large multiline text input field so it is easy to copy and paste to some other window. 

2) A better flag name might be `scenarioLogsInReports`.  This should be at the root level of the backstopjs config.

SeanMcMillan:
I've got the logs capturing onto the filesystem, and I've got them included in the HTML report's config.js. I'm working on making them display in the report, but I haven't worked with Redux before. I'm hoping to have a PR for this either tomorrow or early next week.

evilru:
I don't get scenarioLogsInReports working.

Is there anything I need to do beside configuring it as done in this file: https://github.com/garris/BackstopJS/blob/master/test/configs/backstop.json#L54 ?

I run `backstop remote` and in a second terminal `backstop openReport`.
The approve buttons are visible in the report, but there is no error output from the console.

Best

cojaco:
Hi, same here, backstop is reporting in remote mode. The approve button is visible, but I can't find the log output.

garris:
I will try to look at it this week.   Thanks for reporting.

[...]

PR #1364, opened 2021-10-11
Capture Logs and display them in the reports.

SeanMcMillan:
For #1362.

Please let me know if you want me to change this up in any way and I'll be glad to make the changes.


garris:
Thanks so much.  Will review as soon as I can.  Cheers.

SeanMcMillan:
Have you gotten any chance to look at this? (I don't want to be a nag, I just noticed I was still deploying from my own repo at work and was reminded.)

garris:
Sorry about this -- I did start going through the changes a few weeks ago.   I unfortunately forgot to click the "review" button.  Had an initial comment.

core/util/runPuppet.js
```diff
@@ -262,11 +279,13 @@ async function processScenarioView (scenario, variantOrScenarioLabelSafe, scenar
   if (error) {
     const testPair = engineTools.generateTestPair(config, scenario, viewport, variantOrScenarioLabelSafe, scenarioLabelSafe, 0, `${scenario.selectors.join('__')}`);
     const filePath = config.isReference ? testPair.reference : testPair.test;
+    const logFilePath = config.isReference ? testPair.referenceLog : testPair.testLog;
     testPair.engineErrorMsg = error.message;
 
     compareConfig = {
       testPairs: [testPair]
     };
+    await writeScenarioLogs(config, logFilePath, logger);
```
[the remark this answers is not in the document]
SeanMcMillan: `writeScenarioLogs` only actually writes if the config flag is set. Otherwise it returns a fulfilled promise.

core/command/report.js
```diff
@@ -77,6 +77,29 @@ function writeBrowserReport (config, reporter) {
   logger.log('Writing browser report');
 
   return fs.copy(config.comparePath, toAbsolute(config.html_report)).then(function () {
+    // Slurp in logs
```
garris, line 80: Should this block run if user opts out of log writing?

garris (commented):
Question about appending logs into report datafile.

core/command/report.js
SeanMcMillan: I ran into some problems in testing where the config changes between the reference and test runs, meaning that if the config is set _now_, it doesn't tell us if it was set _then_. So the log manipulating code in this block catches and ignores errors while reading the log files.

We'll need that anyway for the case where there aren't log files as part of the reference, but there are as part of the test. We could skip the block if the config is currently disabled; I'll make that change.

SeanMcMillan:
I've made the change so that it only tries to touch the log files if scenario logs are currently enabled.

garris:
@SeanMcMillan   Thanks for making that change.   Just so I am understanding -- It looks like first we generate log assets.  Then when all tests have completed, we append all the logs to the datafile which is then read in by the client.  Is that correct?

SeanMcMillan:
Yeah.

I was having problems where sometimes the process would end early and not get to report generation, but I still wanted to see the logs to figure out what the problem was, so I wrote them to disk first. (This turned out to be caused by my onReady script throwing a fatal error in certain rare conditions.) Having them in the report is much nicer though, if you get all the way to completion.

garris:
@SeanMcMillan  Thanks for that explanation.  How big are those log files on average per scenario..   EG total size of log directory divided by number of scenarios?

I am wondering if it is actually a better design to keep those log files out of the config file and simply side load those into the report as needed?    IOW: the log asset would behave like more like an IMG asset.   I imagine those things are going to be pretty big -- i am concerned if you have 100 scenarios and each one is .5M the data file will take a lot of compute power to parse.

SeanMcMillan:
About 0.1k a piece, from the smoke test (10033 bytes in 69 files.) If you have more logging, they'll be bigger, of course.

I was initially hoping to demand load the log files, but I couldn't find a way to do so from the local file system. `fetch` and XHR are blocked from loading `file:///` URLs for security reasons.

garris:
If you run `backstop remote` in another terminal it will run a webserver already configured to serve out of your backstop_data directory.   The UI report will automatically detect this and use http calls to resolve assets.  This option also unlocks new features including `approve` right from the browser report.   I recommend you try running this and base your feature on it.   It simply hasn't been documented yet but this is a great opportunity to document this feature.

SeanMcMillan:
I've changed the feature to load the reports via `fetch` and recommend `backstop remote` if the report was loaded via `file:` URL.

Take a look and let me know what you think.

garris:
Wow, sounds great!  On holiday for a week.  Will review as soon as I can!  Happy holidays.

garris (approved):
LGTM.  Will test this out locally.  Thank you for the updates!

garris:
@SeanMcMillan Thanks again for the PR.  I tested this locally and have some more requests.

1. can you move the button in the report.  It currently doesn't follow the design of the rest of the report..

garris:
@SeanMcMillan  Thanks again for the PR. I tested this locally and have 2 more requests...


- Can you update the read me file with instructions on how to use this feature -- including the bit about running `backstop remote`.

- Can you move the button in the report?   It currently doesn't follow the design of the rest of the report -- see the note in the screenshot below...

Screen Shot 2022-01-06 at 5 14 58 PM

SeanMcMillan:
Oops, I didn't mention here that I made PR #1386 for the requested UI changes.

garris:
Thank you for the heads up!  I will take a look as soon as I can.

PR #1368, opened 2021-10-25
Metadata for scenario added for json report

justyna-olszak-wttech:
Project I use BackstopJS in required an option to pass some metadata for each scenario into the final JSON output, to be used by other monitoring tools. The solution is covered in this fork.
Such use case is for example JIRA ticket number. It allows to have additonal information next to tested pair for easier project maintenance and stories reference.

I  believe it is  valuable addiont, that can be used  also by other teams.  @garris

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,7 +9,7 @@ ENV \
 RUN apt-get update && \
 	apt-get install -y git sudo software-properties-common
 
-RUN sudo npm install -g --unsafe-perm=true --allow-root https://github.com/justyna-olszak-wttech/BackstopJS.git#v5.4.5
+RUN sudo npm install -g --unsafe-perm=true --allow-root https://github.com/justyna-olszak-wttech/BackstopJS.git#v6.0.4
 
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -9,7 +9,7 @@ ENV \
 RUN apt-get update && \
 	apt-get install -y git sudo software-properties-common
 
-RUN sudo npm install -g --unsafe-perm=true --allow-root backstopjs@${BACKSTOPJS_VERSION}
+RUN sudo npm install -g --unsafe-perm=true --allow-root https://github.com/justyna-olszak-wttech/BackstopJS.git
```
garris, line 12: Unfortunately we want to stay away from loading private libraries like this.

garris (commented):
Thank you for this PR. I don't think this is something I would want to add to the code base -- but it looks very interesting-- it would probably be more useful to others as a blog post if you can show more about how you will use the metadata in your project.   If you create a post I am happy to add the link to the readme doc.   Cheers!

docker/Dockerfile
justyna-olszak-wttech: Yes, I was pretty sure it will not meet acceptance, but this was necessary to make fork use my code within Docker build.

Issue #1344, opened 2021-07-29
[Found a fix] Problems with usage of puppeteer screenshooting


r-oldenburg:
Hey guys,

I had several hard days of debugging when I found a solution to several randomly appearing problems that we encountered, whenever we switched to a Backstop (docker) version > 4.4.2.

And there are several other issues that (for me) look related to this...

Our problems:
 - Screenshots were taken at the wrong location (having an offset left and top), sometimes with the wrong height
 - Screenshots were different caused by transitions that should not have occured (i.e. hovered elements without any interaction or definition for it, elements moved b/c of an collapes/expand transition that normally should only happen on uer interaction)
 - This was happening **completely unpredictable** on any shot. But mostly happened on large viewport "lg" 1440x768 (we have 4 of them xs,sm,md,lg) and mostly on elements further down the page. Also more on the scenarios that were using `selectorExpansion=true`

Findings from investigation:
 - It turned out that the height offset was somewhat related to the height of images on the page (offset was always a multi)
 - Any try with "waiting for the transitions/animations to end" did not help
 - Also was it clear that we would have two types of problems: 
 - 1. Screenshots with wrong calculation of position/height
 - 2. Screenshots taken while an animation/transition was ongoing

The Fix/Workaround, both within source of backstop:
 - Do not call puppeteers screenshot function in parallel (currently happening via async call in core/util/runPuppet.js:401)
 - Do pass in `captureBeyondViewport: false`
I will add a patch here and if I find the time I can create PR as well.

Also did I have to change our viewport definition to height: 10000 (because `captureBeyondViewport: false` probably would keep us from getting screenshots further down on the page, I suppose. Haven't tested it without)

**THAT FIXED ALL THE PROBLEMS FOR US.**

For over a week now we are working with Backstop 5.3.4 in our CI pipeline (10 Devs, gitlab pipeline, ~10-20 MRs per day) using this patched version and it works just like a charm.

r-oldenburg:
Patch file cannot be attached, so I post it here:

```
From 5e28dd9030b4962245e938c39c133de32af1e592 Mon Sep 17 00:00:00 2001
From: Roland Oldenburg
Date: Wed, 21 Jul 2021 13:58:12 +0200
Subject: [PATCH] Workaround - cropped shots and resize-side-effects


diff --git a/core/util/runPuppet.js b/core/util/runPuppet.js
index 2bbb20f..61cc8c1 100644
--- a/core/util/runPuppet.js
[...]

PR #1374, opened 2021-11-12
add captureBeyondViewport flag for Puppeteer

saitho:
Since v5.3.7 Puppeteer does not capture beyond viewport anymore, which could cause some issues for people relying on it.
A new flag `captureBeyondViewport` for toggling that setting is introduced in engineOptions as discussed in #1344.


saitho:
@garris I was wondering, is there a documentation on available engineOptions?
I wanted to add the new flag but didn't find any documentation on that. :(

garris:
Hi @saitho  -- the `engineOptions` props are pretty much passed directly into puppeteer...
image

For more information about how this can be used you can look at Puppeteer documentation.

garris:
@saitho Does this new config option mean that we do not need this hack anymore?

https://github.com/garris/BackstopJS/pull/1374/commits/e3a80e207088316aa03fcaf70e9322b51a0ec335

saitho:
> @saitho Does this new config option mean that we do not need this hack anymore?

No, it just allows the user to toggle `captureBeyondViewport`.
I want to use BackstopJS to do regression tests on a styleguide generated via kss-node.
I need to set `captureBeyondViewport: true` for it to work properly, as visibility in a viewport is not important here.

Oh I see, `config.engineOptions` is used there. I thought `config` was already the `engineOptions`.
Then my description is wrong. I'll change that. This is not a puppeteer setting, but a new BackstopJS setting.

saitho:
I adjusted the commit message

PR #1377, opened 2021-11-17
Fixed Puppeteer ENOENT Error When Image Stubbing

zachOS-tech:
image

The imageStub.jpg file is only one folder back from the interceptImages.js file. So, I removed the extra `../` so that it can reference the image.

Verified this resolved the Puppeteer ENOENT error message, and Backstop is successfully stubbing images on my computer now.


garris:
Oh great -- thanks for this!

Issue #1142, opened 2020-01-26
Are there plans to support playwright


wesselbaum:
I recently learned about https://github.com/microsoft/playwright which is a project by former puppeteer developers but now working for microsoft instead of goodle and now supporting multiple engines ( Chromium, WebKit and Firefox) wwith a single API. 

Being able to check for Chrome and Firefox would be such a great feature in my opinion.
The API should be similar to puppeteer which is currently implemented.

Any thoughts?

garris:
Yes.  We will absolutely support playwright.  We were just talking about last week.

garris:
If anyone is in a rush -- feel free to add as a new engine and send a PR.

fuhlig:
If it's about Firefox support:
It is supported with pupeeteer 2.1+
https://github.com/puppeteer/puppeteer#q-what-is-the-status-of-cross-browser-support

FireGhost:
I've started an implementation of Playwright: https://github.com/FireGhost/BackstopJS-Playwright

**For now, it does not have all the features that are available for Puppeteer !**

So, the features that works with Playwright right now are:
- viewports
- scenarios (of course)
  - readyEvent
  - readySelector
  - delay
- onReadyScript
- onBeforeScript
- asyncCaptureLimit
- asyncCompareLimit
- misMatchThreshold
- requireSameDimensions
- engineOptions

New features:
- The "browsersTypes" option where you can specify multiple browsers (should work with Puppeteer too). Ex: `"browsersTypes": ["firefox", "chromium", "webkit"]`
- The "hideBrowserConsoleLogs" option, self explained. Set it to "true" in order to hide the console log coming from the browser.

Features that does not work:
- You can only take fullpage screenshot (selectors screenshot is not implemented)
- Reference URL is not implemented
- hideSelector, removeSelector not implemented, but can be done in the onReadyScript
- keyPressSelector, hoverSelector, clickSelector, scrollToSelector not tested/implemented, but can be done in OnReadyScript

masi:
@FireGhost Any updates on this? If you don't want to make a PR, sharing a fork on Github will be appreciated.

garris:
@FireGhost apologies for the delay.  This looks like a great start.  Please do create a PR and let's bring this in.  

We would ultimately want to have parity across engines but it seems ok to start here since this is effectively behind a feature flag.

marcdacz:
[...]

PR #1379, opened 2021-11-18
Playwright fixes: ?? and missing test scripts

FrostyShosty:
Refs #1142 

There are two fixes in this PR:
1) I removed the ?? operator, only supported in ECMAScript 11 for better backwards compatibility.  Sorry about slipping it in there!

2) I copied in the playwright scripts missing from the test case.  Now all smoke tests seem to pass.  I've also updated the reference images for the 'cookies', 'click', and 'keyPressSelector' tests that weren't successfully passing before.


garris (approved):
Looks good.  Thank you for updating the tests!

PR #1395, opened 2022-02-17
Update to latest node LTS: 16.14.0

anthony-hull:
I built this locally and my project's tests still run using the built image after the upgrade.
Is there any other testing you would want doing on this?

It does include an upgrade to a new version of Debian as the node 16 image uses a new major release .

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14.17.3
+FROM node:16.14.0
 
 ARG BACKSTOPJS_VERSION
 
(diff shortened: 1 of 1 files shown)


garris:
Thank you!

fuhlig:
Additionally to the node update there was this change:
https://github.com/anthony-hull/BackstopJS/blob/1dea8826051fc0254ce75d3c8039f321dd90dda7/docker/Dockerfile#L32

Could you please add details what / why this change was made? Was not able to retrace it in the commit
Thanks in advance :)

anthony-hull:
The upgrade to Debian Buster via moving to the new node image has this missing as a dependancy for puppeteer.
Once I added this it worked again.

fuhlig:
Thanks for clarifying 👍

PR #1399, opened 2022-03-06
bugfix(playwright): timeout for elements without height

fuhlig:
If `selectors` targets elements without height (e.g. child elements with `position: absolute|fixed|sticky`),
it times out when using Playwright as the Engine.

Extend global `page.waitForSelector` for `readySelector` with `state: 'attached'` which should ensure that the element is in DOM.

Puppeteer does not have this issue and does not allow this option in `waitForSelector` ([Puppeteer Docs](https://pptr.dev/#?product=Puppeteer&show=api-pagewaitforselectorselector-options))

[Playwright Docs](https://playwright.dev/docs/api/class-page#page-wait-for-selector)


fuhlig:
`state` is not safe.
Needs different handling for opting to `fullPage` screenshot when element has no height.

PR #1403, opened 2022-03-19
chore: update puppeteer from 12.0.1 to 13.5.1

LeonEck:
The version of puppeteer currently bundled with Backstop depends on a vulnerable version of node-fetch. This PR bumps the version of puppeteer to the latest one to get rid of the vulnerability.

I am not sure if the smoke tests are working correctly for me. After cloning the repo without making any changes I already get 32 failed tests with smoke-test-docker. For some of them, it is obvious that they can't work, since the reference images don't have "Playwright" in the first paragraph. Many other tests fail with diff-y plus or minus 1.

At least the results didn't change after I updated puppeteer. I attached the report from smoke-test-docker to this PR: [BackstopJS Report.pdf](https://github.com/garris/BackstopJS/files/8309445/BackstopJS.Report.pdf)


garris:
Thank you for this.  Yes, the tests are just catching a text change.  Will update this before publishing to npm.

PR #1404, opened 2022-03-19
chore: update playwright from 1.17.1 to 1.20.0

LeonEck:
Thank you for merging https://github.com/garris/BackstopJS/pull/1403. This PR includes another dependency update that would be much appreciated.

In one of my projects I use playwright for my E2E tests and as an engine in backstop. With the latest release of playwright the version difference between 1.20 (the latest) and the bundled backstop version (1.17.1) seems to be too large. After using playwright 1.20 to install the browsers, backstops playwright 1.17.1 can't find them.

I tried out this PR in my project and it resolves this issue.


garris:
Thanks.

PR #1417, opened 2022-05-05
Reuse playwright browser

FrostyShosty:
This is a possible fix to #1416.  Using James Irish's idea to reuse the playwright browser, this code only creates the browser once, and creates independent browser contexts for each scenario, safely closing the browser at the end.

Passes smoketests (relatively?) quickly, but I do not have his 624-scenario setup to truly test the speed.

PR #1419, opened 2022-05-10
Quick fix for early disposal of playwright browser

JamesIrish:
Pass the browserContext instead of the browser instance.  See comment https://github.com/garris/BackstopJS/pull/1417#issuecomment-1121303500


garris:
👍 thanks

PR #1423, opened 2022-06-24
Document how to suppress firewall popups on macOS

deviantintegral:
It's kind of painful trying to watch your browser when 10s of popups keep spawning!


garris (approved):

garris:
Thank you!

Issue #1158, opened 2020-03-04
Remote Service Documentation


fuhlig:
Currently there is no documentation on the remote service, despite being available via the CLI. While there is active work on this (e.g. approve button, which sound really great), I can't really try it out, give feedback and potentially contribute.

- document 
  - purpose
  - workflow
  - limits
  - usage in CI environment
- configure port ([open issue @ ember-backstop](https://github.com/garris/ember-backstop/issues/34) & #1150)

If you require assistence, I'd be happy to contribute.
Thanks in advance.

garris:
OMG you are so right.  Thanks for getting me to focus on this for 10 minutes.  I will publish this to the readme soon...

# Backstop-Remote

## Description:

Backstop-Remote exposes BackstopJS CLI commands as an HTTP Web service, enabling browser-UI client interactions (e.g. hosted backstop reports, approve button interactions, etc.) with a local or remote instance of BackstopJS. Additionally, Backstop-Remote provides an interface for BackstopJS to interact as a slave web-app screenshot renderer and comparison engine for other JS-based testing and or development frameworks (e.g. Ember, Testem, Cypress, Storybook, etc.)

## Workflow:

Starting the Backstop-Remote service

### On the CLI...

```
backstop remote --config=<optional-some-backstop-config>
```

### In a node app you might do something like this...

```
    backstopjs('remote', {
      config: commandOptions.config, 
      filter: commandOptions.filter
    });
```

## Stopping the service

Doesn't get more restful than this...

```
http://<backstopHost>/stop/
```

## Usage in CI env

Just start the service and ensure the host name and port are hit able inside your CI boundary.


## For more info...

See: https://github.com/garris/ember-backstop

Or you can bug me, Garris! 
@garris on twitter 
or find me on linkedin.

WBTMagnum:
I'm fiddling around with the remote feature using the docker container. I'm able  to execute `/stop` and `/version`, but I cannot figure out who to access or open the report for the given configuration.

I also had a look at [/remote/index.js](https://github.com/garris/BackstopJS/blob/master/remote/index.js) and tried to use `/dview/:testId/:scenarioId`, but to no avail. I always get the response "Cannot GET /" (respectively "Cannot GET /<given-path>").

Any hints on how to create and/or access the report?

dgrebb:
@WBTMagnum try [http://localhost:3000/backstop_data/html_report/index.html](http://localhost:3000/backstop_data/html_report/index.html)

[...]

PR #1445, opened 2022-10-31
Remote docs

domibarton:
I had some issues to setup the remote server, resp. I didn't know how to get the interactive button shown at first. Thus I did some reverse engineering and got it working. I always tried to browse to the page directly (without explicitly opening the report), which of course just showed me static site without approval button. Because of that (and issue #1158), I thought I'll updated the docs. Let me know if this is OK or if you want some changes.


garris:
Thank you!

Issue #1466, opened 2023-03-24
Failing to create bitmaps_test folder in GitHub Actions after upgrade to 6.2.0


davidcmoulton:
We have successfully been using `backstopjs` in our GitHub Actions with 6.1.4, using `npx backstop --docker test`. On upgrading to 6.2.0 the backstop run in GitHub Actions fails, erroring with a bunch of:
```
Browser Console Log 0: JSHandle:BackstopTools have been installed.
Error: EACCES: permission denied, mkdir 'backstop_data/bitmaps_test'
```
We suspect the regression may be stemming from the changes in user introduced in https://github.com/garris/BackstopJS/commit/6d221a275686bad7c7b58fd7ff1325e7067e1bcb.

garris:
@tkrah 👆 any thoughts on this one?

tkrah:
Sorry, with that few input, can't really comment on that, I don't use GitHub actions, so can't comment on those either.
I use that image of 6.2.0 without a problem, eaccess looks like a permission problem though, op should investigate in that direction.

garris:
Thank you @tkrah for the fast enhancement!   

@davidcmoulton can you please validate that this patch solved your issue?

erkannt:
I'm part of the same ensemble as @davidcmoulton and can confirm that the patch works for us:

https://github.com/sciety/sciety/pull/2168

theghostbel:
Does it work when running from Windows?

tkrah:
> Does it work when running from Windows?

depends - if you have an environment where id is available yes, if not than you need to customize the dockerCommandTemplate like written in the README.md of backstopjs to use one which does work for your environment.

tkrah:
You need to use a username which your docker daemon in windows does know about. As I don't know your setup, you have to find out yourself which username you need to provide. You can also try "--user root" but that will get you files owned by root - which may or may not be a problem for you to be removed later because of the ownership, ymmv.

theghostbel:
ok, `--user root` works well, no problems so far. Should add this to docs, quite a tricky problem to investigate. 

```
dockerCommandTemplate: 'docker run --rm -it --user root --mount type=bind,source="{cwd}",target=/src backstopjs/backstopjs:{version} {backstopCommand} {args}'
```

tkrah:
@garris Your code, your decision ;-). 

[...]

PR #1467, opened 2023-03-24
fix permission problem with local bind mount (#1466)

tkrah:
npm run sanity-test-docker
npm run smoke-test-docker

both are able to run now without permission problems.


tkrah:
I did not catch that problem initially, because I am already running the backstopjs container with my own user via "--user" command because root owned where a mess ;).

garris (approved):
LGTM

Thanks @tkrah!

PR #1471, opened 2023-03-31
build (engines)!: Update dependencies for engines

fuhlig:
- update `puppeteer` and `playwright`
- updates browser versions:
  - Chrome: `104` => `112`

BREAKING CHANGE:
run `npm install`

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,9 +103,9 @@
     "os": "^0.1.2",
     "p-map": "^4.0.0",
     "path": "^0.12.7",
-    "playwright": "^1.23.4",
+    "playwright": "^1.32.1",
     "portfinder": "^1.0.32",
-    "puppeteer": "^15.4.0",
+    "puppeteer": "^19.8.2",
     "super-simple-web-server": "^1.1.3",
     "temp": "^0.9.4"
   }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,290 +1,608 @@
 {
   "name": "backstopjs",
   "version": "6.2.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
   "requires": true,
-  "dependencies": {
-    "@babel/code-frame": {
+  "packages": {
+    "": {
+      "name": "backstopjs",
+      "version": "6.2.0",
+      "license": "MIT",
+      "dependencies": {
+        "@mirzazeyrek/node-resemble-js": "^1.2.1",
+        "chalk": "^4.1.2",
+        "diverged": "^0.1.3",
+        "fs-extra": "^10.1.0",
+        "jump.js": "^1.0.2",
+        "junit-report-builder": "^3.0.0",
+        "lodash": "^4.17.21",
+        "minimist": "^1.2.6",
+        "object-hash": "3.0.0",
+        "opn": "^5.5.0",
+        "os": "^0.1.2",
+        "p-map": "^4.0.0",
+        "path": "^0.12.7",
+        "playwright": "^1.32.1",
+        "portfinder": "^1.0.32",
+        "puppeteer": "^19.8.2",
+        "super-simple-web-server": "^1.1.3",
+        "temp": "^0.9.4"
+      },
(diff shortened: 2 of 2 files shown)


garris (commented):
Thanks so much for the update.  Do you mind adding a link to the breaking changes?

tkrah:
You should consider that your docker image does only support chromium up-to version:

108.0.5359.94-1~deb11u1

I am not an expert here if puppeteer with chrome now on 112 will function for sure - maybe use a puppeteer version which supports 108.

package.json
```diff
@@ -103,9 +103,9 @@
     "os": "^0.1.2",
     "p-map": "^4.0.0",
     "path": "^0.12.7",
-    "playwright": "^1.23.4",
+    "playwright": "^1.32.1",
     "portfinder": "^1.0.32",
-    "puppeteer": "^15.4.0",
+    "puppeteer": "^19.8.2",
```
tkrah, line 108: v19.2.0 to match the provided chromium version in the docker image.

tkrah (changes requested):
please change the puppeteer version.

tkrah:
We could use the chromium from the playwright install which is 112.0.5615.29 - but I am unsure how to *know* the path for that installation, because it has a version component in it and from the playwright docs I don't see a way to specify a concrete version, so that we can have a stable working build and set the puppeteer variable to the right path. Ideas anyone with more knowledge of playwright?

fuhlig:
Could you elaborate on the compat issues?

Running backstop with `puppeteer@19.8.2` (Chrome 112) works when running with `--docker`.

When building a Docker image with `node:16-bullseye` (which could be updated to v18 LTS) and running `apt-get install chromium` I get
```
Chromium 111.0.5563.110 built on Debian 11.6, running on Debian 11.6
```

tkrah:
It may work, but it is not guaranteed and issues are likely according to puppeteers dev people. So better be safe here and use a version which does match.

https://packages.debian.org/search?suite=bullseye&searchon=names&keywords=Chromium does only result in 108.x ... Interesting where that 111 is coming from.
But no matter what, please use a puppeteer which does match that 111 than. Need to have a look tomorrow where 111 is coming from.

fuhlig:
@tkrah 
Thanks for checking out and clarifying.

garris:
Thank you @fuhlig and @tkrah!

fuhlig:
Cheers for the quick responses and merge! 🚀 

When is the next npm release planned?

garris:
Hi.  I'm still traveling.  Will be back Saturday -- will try to do it on the weekend if I can get a few mins.  

Cheers

garris:
@tkrah @fuhlig  BackstopJS v6.2.1 is released to NPM & Docker.   Thanks for your patience.  Cheers.

tkrah:
Puppeteer versions must match certain chrome versions, as the base image demands a certain chrome version we have to migrate anyway, no matter if it breaks or not if we want to ensure that puppeteer is compatible (according to their docs) - so it was already a "bug" in backstopjs to not match those demands in the past - so fixing bugs even if they break things is ok for patch versions. @genepaul  - so yes your changes may be coming from there (you can easily test it by reverting that change and rebuild backstopjs) - but you have to cope with the changes anyway - you can stay on an older version if you can't upgrade yet.

genepaul:
@tkrah you and I have different understandings of how semantic versioning works. A "fix" that breaks things is still a breaking change, thus a major version, which would allow consumers to _opt in_ to the fix. Releasing as a patch forces those who may have used semver ranges to pull changes in without realizing it.

tkrah:
You are right, but no where did I found in the docs of BackstopJS that it does follow semantic versioning, did you? So I would not expect it to follow those rules if the author does not told us to do so.

genepaul:
lol ok. I guess I thought setting the version numbers to <major>.<minor>.<patch> was an implicit indication that this library followed a well-known practice in the open source community. My bad for making such an assumption.

Issue #1468, opened 2023-03-27
Chromium not available for non root user within docker for Playwright at version 6.2.0


websi:
With Version 6.2.0 the chromium browser is not available for non root user. The normal user is `node`. See https://github.com/garris/BackstopJS/blob/1144782deebfe7e70650f09cc7fa2d3ffb349f7a/docker/Dockerfile#L18

```
$ docker run --rm -v $CI_PROJECT_DIR:/src  backstopjs/backstopjs:6.2.0 test --config=backstop.config.js
BackstopJS v6.2.0
Loading config:  /src/backstop.config.js 
COMMAND | Executing core for "test"
createBitmaps | Selected 10 of 10 scenarios.
Creating Browser
      COMMAND | Command "test" ended with an error after [0.007s]
      COMMAND | browserType.launch: Executable doesn't exist at /home/node/.cache/ms-playwright/chromium-1048/chrome-linux/chrome
                ╔═════════════════════════════════════════════════════════════════════════╗
                ║ Looks like Playwright Test or Playwright was just installed or updated. ║
                ║ Please run the following command to download new browsers:              ║
                ║                                                                         ║
                ║     npx playwright install                                              ║
                ║                                                                         ║
                ║ <3 Playwright Team                                                      ║
                ╚═════════════════════════════════════════════════════════════════════════╝
                    at module.exports.createPlaywrightBrowser (/usr/local/lib/node_modules/backstopjs/core/util/runPlaywright.js:43:42)
                    at /usr/local/lib/node_modules/backstopjs/core/util/createBitmaps.js:136:7
                    at delegateScenarios (/usr/local/lib/node_modules/backstopjs/core/util/createBitmaps.js:1[35](https://gitlab.dkd.de/dkd/products/typo3-cms-standard/-/jobs/242187#L35):12)
                    at module.exports (/usr/local/lib/node_modules/backstopjs/core/util/createBitmaps.js:192:19)
                    at Object.execute (/usr/local/lib/node_modules/backstopjs/core/command/test.js:17:14)
                    at Object.execute [as test] (/usr/local/lib/node_modules/backstopjs/core/command/index.js:59:49)
                    at execute (/usr/local/lib/node_modules/backstopjs/core/command/index.js:109:31)
                    at module.exports (/usr/local/lib/node_modules/backstopjs/core/runner.js:6:10)
[...]

PR #1472, opened 2023-03-31
use shared playwright browser location path (#1468)

tkrah:
@garris another PR for you to consider ;-) .


tkrah:
@tsvetislavt99 Can you try that fix?

garris:
Thanks @tkrah!

PR #1476, opened 2023-04-19
Rollback docker user change

tkrah:
- defaults to root now again
- added a recommendation to the README.md to configure a more appropriate user to run the test by changing the dockerCommandTemplate config option
- fixes (#1466)

### DIFF
--- a/README.md
+++ b/README.md
@@ -690,6 +690,12 @@ The above flag will cause BackstopJS to hit your Docker local client, spin up a 
 If the default docker command or image does not work for you, you can customize the command to run BackstopJS with Docker by changing the `dockerCommandTemplate` config option. The default is:
 
 ```sh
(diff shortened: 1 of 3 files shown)


tkrah:
@garris That would be my proposal like discussed - nothing better came to my mind to honor your expectations ;-)

garris (commented):
This looks great!

neclimdul:
@garris this looks like the fix to something I was running into but it doesn't seem to have made its way into a tagged release on over to docker hub. any changes you could do that?

garris:
Will look at this today

garris:
Yup -- looks like the last publish didn't work.   Just pushed a new version please LMK if it's not working as expected.

Cheers.

neclimdul:
Looks great! Thanks!

Issue #1485, opened 2023-06-05
Puppeteer old Headless deprecation warning


neclimdul:
Seeing a bunch of these in the latest release on docker hub.

```
  Puppeteer old Headless deprecation warning:
    In the near feature `headless: true` will default to the new Headless mode
    for Chrome instead of the old Headless implementation. For more
    information, please see https://developer.chrome.com/articles/new-headless/.
    Consider opting in early by passing `headless: "new"` to `puppeteer.launch()`
    If you encounter any bugs, please report them to https://github.com/puppeteer/puppeteer/issues/new/choose.
```

Seems mostly informational and not problematic but there's just a swarm of them when testing. Quick skim looks to be caused by the `headless: <!!!config.debugWindow>` logic which passes true. The boolean cast also means we can't put 'new' in the config and "opt in" so I haven't tested if opting in is functional or not.

garris:
Thanks for raising!  Looks like we should update this behavior to allow updating with non-bool values and to make a decision on how backstop should interpret the old backstop default value going forward.   Would be happy to review a PR on this.  Cheers

tkrah:
The new mode however should be treated with care - e.g. https://bugs.chromium.org/p/chromium/issues/detail?id=1440046 the performance loss is extreme and at least for me in docker I can't even use the new mode because angle fails to initialize the egl rendering acceleration and just quits, I am forced at the moment to use the "old" one and I guess I am not alone with that. So if the new one is considered make it configurable for the user and better stay on the old one for the moment.

alexd3d:
Guys, thanks a lot for great tool!
Just came across this issue too. If we won't switch to new headless, it would be great to have some additional config param for BackstopJS that regulates verbosity of console. I have 41 test and console looks really scarry with all that millions lines that are logged. Would be great that we could at least : 1) turn on\off warnings yellow about headless 2) turn on\off original tested website's console messages. Maybe it's possible now but I couldn't find? Thanks!

dgrebb:
@alexd3d FYA hoping to get this merged in soon: #1511 

@garris this is a super-small adjustment which should work in all cases backwards and forwards :)

dgrebb:
Another update on this with PR #1535 1535

Playwright is going to stay defaulted to "old" headless mode for now, but Puppeteer will start default to `new` if/when #1535 is merged.

@tkrah I included a fallback for "old" in Puppeteer, which, ironically, can be set by passing `"headless": true` in the config.

PR #1486, opened 2023-06-22
potential fix for puppet warning

Neychok:
Would something like this fix https://github.com/garris/BackstopJS/issues/1485 ?

I've never contributed to BackstopJS before so I am sorry if I missed any steps when creating this PR.

Happy to work on this more, if any other changes are required.

### DIFF
--- a/core/util/runPlaywright.js
+++ b/core/util/runPlaywright.js
@@ -33,10 +33,12 @@ module.exports.createPlaywrightBrowser = async function (config) {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


tkrah:
You would need to pass the string "new" to fix that warning - but I can't recommend using new yet, see the issue comment.

dgrebb:
Thanks for the work on this @Neychok — I submitted a new PR which will work with backwards compatibility in mind.

PR #1489, opened 2023-07-30
fix(remote): allows the `ApproveButton` component to get its port num…

dgrebb:
…ber from the browser

currently this component is trying to read from `process.env.BACKSTOP_REMOTE_HTTP_PORT`, which the browser has no access to. not sure if the intent is for webpack to interpolate this value and output a string, but it currently isn't. let the browser do the work for you :)


dgrebb:
@garris tagging you here as I think this will help quiet down some additional issues now that `remote` may be back in business.

garris:
Thanks @dgrebb!

PR #1500, opened 2023-09-21
Feature/indexed testid report json

dgrebb:
This comes from the need to include `index` in the ouptut of automation for dgrebb.com. Using the `index`, we'll be able to link directly to a test in the report via its `id` HTML attribute.

This will be used in GitHub workflow output to quickly link to the failing report on repots.dgrebb.com


dgrebb:
mistaken PR

PR #1511, opened 2023-11-22
fix: adds headless ternary to handle 'new', true, and false in a non-…

dgrebb:
…breaking fashion (#1485) (#1486)

_Adapted from conversation in #1485 and @Neychok's initial PR in #1486._

## Summary

Puppet and Playwright complain and pollute logs with warnings about "new headless mode". This reads the value set in backstop.json, or uses true if 'headless' is not set, or 'new', as a string. README updated with details.

## Tests

Updated all `./test/configs` with `headless: 'new'` and ran sanity tests. For example:

```shell
File: test/configs/backstop.json
46:   "engineOptions": {
47:     "args": ["--no-sandbox"],
48:     "headless": "new"
49:   },
```

Closes: #1485

### DIFF
--- a/README.md
+++ b/README.md
@@ -644,12 +644,13 @@ ignoreHTTPSErrors: true,
 headless: <!!!config.debugWindow>
 ```
 
-You can add more settings (or override the defaults) with the engineOptions property. (properties are merged)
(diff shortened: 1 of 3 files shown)


garris:
Thank you @dgrebb!

PR #1530, opened 2023-12-23
Bugfix/1529 font family regression

### LINKED ISSUE #1529: UI style regression with 6.3.2
Hey @dgrebb -- any hunches where this issue may have crept in?

image

_And FWIW -- it's not lost on me that backstop UI doesn't have any UI Regression tests.  Proof that Irony is not dead.  😄_

dgrebb:
The fix can be attributed to a missing colon — seen with the introduction of `TextDetails.js`. 

This goes all the way back to [3.1.15](https://github.com/garris/BackstopJS/blob/v3.1.15/compare/src/components/atoms/TextDetails.js#L23)!

Closes #1529


garris:
Thank you very much for this fix!

PR #1532, opened 2023-12-23
Feature/GitHub workflow build compare

dgrebb:
[Build Compare Test](https://github.com/dgrebb/BackstopJS/actions/workflows/build-compare.yml)

@garris ghis adds a test to see if `build-compare` compiles successfully.

I also drastically reduced the `git checkout` time by reducing the depth to `1`, which is the current commit only, and no further history.


garris:
Thank you @dgrebb!

Issue #1534, opened 2023-12-24
[BUG]: recently added support for `"new"` headless mode fails with Playwright


dgrebb:
## Summary

As documented in [the README](https://github.com/dgrebb/BackstopJS#setting-puppeteer-and-playwright-option-flags), setting `engineOptions.headless = "new"` is passed to Playwright as expected, but fails.

### `backstopjs@6.3.2`

```shell
Creating Browser
No Playwright browser specified, assuming Chromium.
      COMMAND | Command "test" ended with an error after [0.003s]
      COMMAND | browserType.launch: headless: expected boolean, got string
```

## Details

Playwright supports `"new"` headless mode, set as a string, but only by defining `ignoredDefaultArgs: ["--headless"]`.

This isn't well documented, but one can see Playwright's [`headless` argument](https://playwright.dev/docs/api/class-browsertype#browser-type-launch-option-headless) expects a boolean.

Deeply burred in [a GitHub issue](https://github.com/microsoft/playwright/issues/21194#issuecomment-1444276676), there is a small example of how to make `"new"` headless mode work:

```javascript
  projects: [
    {
      name: 'chromium',
      use: {
        launchOptions: {
          ignoreDefaultArgs: ['--headless'],
          args: ['--headless=new']
        }
      },
    },
```

(#1525) (dgrebb/backstopjs#73)

dgrebb:
Found and fixing 👍

dgrebb:
Closed with #1535

Issue #1525, opened 2023-12-21
Headless mode: deprecation warning after upgrading 


fuhlig:
After the upgraded Puppeteer dependency with the latest `6.3.1` release, Puppeteer/Chrome throws the following deprecation warning:
![image](https://github.com/garris/BackstopJS/assets/1364116/cc10e616-5682-47f8-80a8-11afc25786ca)

docs:
https://developer.chrome.com/docs/chromium/new-headless?hl=en

While this is documented in the [Backstop docs](https://github.com/garris/BackstopJS?tab=readme-ov-file#setting-puppeteer-and-playwright-option-flags), it might be a better option is setting the headless mode to `new` by default to avoid deprecation warnings (and potential failures in future versions).

dgrebb:
@fuhlig I agree, and know this is planned for a future release. "new" is not new so much anymore ¯\_(ツ)_/¯

I'll look into submitting a PR for the change today.

fuhlig:
Cheers mate

dgrebb:
@fuhlig I just remembered to follow up! This can probably be closed — check out #1535 and `backstop@latest` ([6.3.3](https://www.npmjs.com/package/backstopjs)).

There are some notes in [the PR](https://github.com/garris/BackstopJS/pull/1535) about Playwright to be aware of. We didn't think it was a good idea to default _that_ runner to `'new'` just yet.

Happy new year!

fuhlig:
Awesome, thanks @dgrebb!
Will check it out next week

Have a happy new year!

PR #1535, opened 2023-12-24
PATCH: Feature/1525 headless mode deprecation

dgrebb:
# Summary

Puppeteer defaults to `"new"` headless mode, as per the feature request in #1525, also discussed in #1485. 

Playwright, on the other hand, will remain `headless: true` by default, as `"new"` mode is not supported without some configuration overhead. And speaking of Playwright...

> [!IMPORTANT]
> #1534 is fixed, which I found while exploring the above-mentioned caveats of `"new"` mode. 
> **This should be considered for an immediate patch**

In the end, we support `"new"` headless mode in Playwright, but only when set by a Backstop configuration file.

I also addressed a fix for GitHub Actions `npm run integration-test` error: `Error: Process completed with exit code 254.`.

Any and all input welcome! We want to be safe, but at the same time, want to support the cutting-edge and standards for Playwright, Puppeteer, and the various browser vendors and teams.

# Details

## Puppeteer

- Puppeteer now defaults to `"new"` headless mode
	- tested docker
	- tested CLI
	- ran through GH actions

## Playwright

From what I've gathered, Playwright is not planning to support `"new"` mode at this point in time. There are performance considerations for their users, so they make it more challenging to use the new mode.

Once Playwright supports `"new"` mode, we should discuss enforcing this default for Playwright, but until then, this is a user configuration.

### Problem

Playwright expects a [boolean for headless mode](https://playwright.dev/docs/api/class-browsertype#browser-type-launch-option-headless).

### Solution

Override Playwrights default args.

We evaluate whether a user is trying to pass `headless` as a string, and include `ignoreDefaultArgs: ["--headless"]`, which allows a string to be passed.

We set the property `headless: true` if `"new"` is passed, and include `--headless=new` in the final `playwrightArgs` sent to the runner.

```javascript
{
 args: [ '--no-sandbox', '--headless=new' ],
 headless: true,
 ignoreDefaultArgs: [ '--headless' ]
}
```

## Notes

- Specific to testing Chrome extensions, the Playwright team as of June 1st, 2023 still recommends "[the normal old headless mode](https://github.com/microsoft/playwright/issues/23389#issuecomment-1571721631)"

### Example Playwright `headless: 'new'` error

```shell
Creating Browser
No Playwright browser specified, assuming Chromium.
COMMAND | Command "test" ended with an error after [0.004s]
COMMAND | browserType.launch: headless: expected boolean, got string
```

### Extra

I started adding inline documentation, which are a nice addition to overall developer experience. The workarounds for Playwright `"new"` headless mode are commented on for future reference.

Pasted image 20231224000613

### DIFF
--- a/README.md
+++ b/README.md
@@ -669,7 +669,10 @@ ignoreHTTPSErrors: true,
 headless: <!!!config.debugWindow>
 ```
 
-You can add more settings (or override the defaults) with the engineOptions property. (properties are merged). This is where headless mode can also be set to 'new', until "new headless mode" is the default in Puppet/Playwright.
+You can add more settings (or override the defaults) with the `engineOptions` property. (properties are merged). This is where headless mode can also be set to 'new', until "new headless mode" is less hacky and more supported by Playwright.
+
+> [!INFORMATION]
+> Puppeteer now runs in `new` headless mode by default, but can be set to `old` headless by passing `"headless": true` in the configuration file.
 
 ```json
 "engineOptions": {
--- a/core/util/runPlaywright.js
+++ b/core/util/runPlaywright.js
@@ -20,25 +20,86 @@ const DOCUMENT_SELECTOR = 'document';
 const NOCLIP_SELECTOR = 'body:noclip';
 const VIEWPORT_SELECTOR = 'viewport';
 
+/**
+ * @method createPlaywrightBrowser
+ * @function createPlaywrightBrowser
+ * @description Take configuration arguments, sanitize, and create a Playwright browser.
+ * @date 12/23/2023 - 10:13:35 PM
+ *
+ * @async
+ * @param {Object} config
+ * @returns {import('playwright').Browser}
+ */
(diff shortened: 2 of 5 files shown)


dgrebb:
@garris this address `new` headless mode. We can default Playwright to use `new` if that is the goal, but because it's a little hacky, I didn't do so (yet).

Please let me know if you'd like an `npm` script and GitHub Action test for this as well. I think it's worth it but didn't want to overload the PR more than it already may be.

garris:
@dgrebb Thanks Dan for doing the extra research on this.  It sounds like you are setting defaults that are in-line with the philosophies of the different frameworks -- which makes sense.

I don't think this new/old/true/false oddness needs to be elevated to a GH test -- I'd prefer to keep GH testing space for high-level critical-path stuff e.g. sanity, smoke, integration, basic UI backstop testing with your cool fixtures idea.   

Thank you!

Issue #1553, opened 2024-02-26
This update broke delay functionality.


garris:
This update broke delay functionality. 
https://github.com/garris/BackstopJS/blob/master/core/util/runPuppet.js#L178
This line is now giving errors.

Puppeteer had breaking changes reported in their release notes: 
https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-core-v22.0.0

I guess that example scripts in init also need updating.

_Originally posted by @furai in https://github.com/garris/BackstopJS/pull/1550#issuecomment-1964621218_

Puppeteer *Delay* feature is no longer supported.

If someone would submit a solution as a backfill (e.g. the promise timeout workaround mentioned here https://github.com/puppeteer/puppeteer/pull/11780) I would gladly publish this backward compatibility fix!

PR #1554, opened 2024-02-27
Change puppeteer's waitForTimeout into direct setTimeout

maciej-siola:
Small fix to address #1553 

In puppeteer waitForTimeout was implemented this way:

```javascript
  /**
   * @deprecated Replace with `new Promise(r => setTimeout(r, milliseconds));`.
   *
   * Causes your script to wait for the given number of milliseconds.
   *
   * @remarks
   * It's generally recommended to not wait for a number of seconds, but instead
   * use {@link Frame.waitForSelector}, {@link Frame.waitForXPath} or
   * {@link Frame.waitForFunction} to wait for exactly the conditions you want.
   *
   * @example
   *
   * Wait for 1 second:
   *
   * ```ts
   * await frame.waitForTimeout(1000);
   * ```
   *
   * @param milliseconds - the number of milliseconds to wait.
   */
  async waitForTimeout(milliseconds: number): Promise<void> {
    return await new Promise(resolve => {
      setTimeout(resolve, milliseconds);
    });
  }
``` 

  
  Therefore it should be sufficient to simply replace function calls with direct setTimeout.

### DIFF
--- a/capture/engine_scripts/puppet/clickAndHoverHelper.js
+++ b/capture/engine_scripts/puppet/clickAndHoverHelper.js
@@ -27,7 +27,9 @@ module.exports = async (page, scenario) => {
   }
 
   if (postInteractionWait) {
-    await page.waitForTimeout(postInteractionWait);
+    await new Promise(resolve => {
+      setTimeout(resolve, postInteractionWait);
+    });
   }
 
   if (scrollToSelector) {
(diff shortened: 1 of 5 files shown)


garris (approved):
Thank you!

furai:
Thanks @maciej-siola & @garris for taking care of this! When can we expect a new release with the fix?

garris:
6.3.21 was just pushed to NPM & Docker Hub.   Thanks to everyone who helped get this patched!

Issue #1555, opened 2024-02-28
`scenario.misMatchThreshold` is ignored


gruberro:
Since using BackstopJS `>=6.3.19` all scenarios using a custom `misMatchThreshold` are ignoring this configuration value. I may be wrong @dgrebb, but IMHO https://github.com/garris/BackstopJS/pull/1539/files#diff-60b20548efe794ef5ed3fb10ae1b2d2fbf9ad3786e7c69d86da083c1b9cd9782R9 introduced this problem by looking for `misMatchThresholdconfig` instead of `misMatchThreshold` 😄.

dgrebb:
Thank you @gruberro for finding this — submitted  #1556 to fix.

gruberro:
> Thank you @gruberro for finding this — submitted #1556 to fix.

Thank you for the immediate fix!

PR #1539, opened 2024-01-23
Feat/1432 global scenario settings

dgrebb:
# Problem

As described in #1432 and other issues, setting/getting  `scenario` properties at the global level would save time, typing, and testing overhead.

# Solution

Spread (`...`) and **logical OR** (`||`) operator.

> [!IMPORTANT]
> This feature is opt-in only, and documented as such

Because configuration  at the `scenario` level will always override any `globals` set in `backstop.json`, a user will need to remove the corresponding property in a `scenario` object.

This is a 100% non-breaking change, thanks to spread`...`
## Supported Global Properties

- `cookiePath`
- `url` - maybe we have a single-page app with no routes?
- `readySelector`
- `delay`
- `hideSelectors`
- `removeSelectors`
- `hoverSelector`
- `clickSelector`
- `postInteractionWait`
- `selectors`
- `selectorExpansion`
- `misMatchThreshold` - previously supported but refactored `getMisMatchThreshHold`
- `requireSameDimensions` - previously supported but refactored `getRequireSameDimensions`

Added to both Playwright and Puppeteer.
# Closes

- [Allow to set scenario option defaults globally](https://github.com/garris/BackstopJS/issues/1432)
- [Global hideSelectors #1309](https://github.com/garris/BackstopJS/issues/1309)
- [Can you use `removeSelectors` globally? #1090](https://github.com/garris/BackstopJS/issues/1090)
- [Chromy: Global `hideSelectors` not working #715](https://github.com/garris/BackstopJS/issues/715)

# Notes

An example:

<details>
<summary>Expand `backstop.json`</summary>

```json
{
  "id": "backstop_playwright",
  "viewports": [
    {
      "label": "phone",
      "width": 320,
      "height": 480
    },
    {
      "label": "tablet",
      "width": 1024,
      "height": 768
    }
  ],
  "onBeforeScript": "playwright/onBefore.js",
  "onReadyScript": "playwright/onReady.js",
  "globals": {
    "cookiePath": "backstop_data/engine_scripts/cookies.json",
    "url": "https://garris.github.io/BackstopJS/",
    "readySelector": "",
    "delay": 0,
    "hideSelectors": [".getItBlock"],
    "removeSelectors": [".logoBlock"],
    "hoverSelector": "",
    "clickSelector": "",
    "postInteractionWait": 1000,
    "selectors": [],
    "selectorExpansion": true,
    "misMatchThreshold" : 0.1,
    "requireSameDimensions": true
  },
  "scenarios": [
    {
[...]

PR #1556, opened 2024-02-28
fix: remove "`config`" from `mismatchThreshold` (#1555)

dgrebb:
As reported in #1555, `mismatchThreshold` was mistyped while introducing scenario config defaults in #1539.

@garris this PR fixes the error of my ways...


garris (approved):

garris:
Thank you @dgrebb!!

Issue #1492, opened 2023-08-07
Playwright runner doesn't support `scenarioLogsInReports`


dgrebb:
Currently Playwright is not logging scenario test `json` files. 

We shall make it so!

dgrebb:
All set with #1493.

Also added `--colors` flag to `remote` `child_process.exec` so we can have colors everywhere :)

@garris hope this helps!

PR #1558, opened 2024-03-05
fix: `misMatchThreshold` and `requireSameDimensions` nullish coalescing (#1557)

### LINKED ISSUE #1557: requireSameDimensions property is not working
According to the documentation, we can set the 'requireSameDimensions' property to 'false' if we do not want to trigger a failed test when the dimensions are not equal. However, after debugging, I discovered that the 'requireSameDimensions' property will never receive a value of 'false' because it is being ignored in this condition. https://github.com/garris/BackstopJS/blob/a0a9507e723d7ab90cc5c698c407b3707198b0d5/core/util/engineTools.js#L46

dgrebb:
This fixes #1557, which I introduced when working on #1492. My bad!

### DIFF
--- a/core/util/engineTools.js
+++ b/core/util/engineTools.js
@@ -6,7 +6,7 @@
  * @returns {number} The mismatch threshold value.
  */
 function getMisMatchThreshHold (scenario, config) {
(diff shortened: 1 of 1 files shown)


garris:
Thank you @dgrebb!   Will push tomorrow.  Cheers!

Issue #1482, opened 2023-05-17
(Feature) Use backstop.js instead of backstop.json by default


klodoma:
Hi @garris 

Did you ever considered using a backstop.js(by default) instead of the backstop.json. 

I know one can use it with `backstop test --docker --config=backstop.js` but I think it would be more efficient for the future. 
I use this for all projects where I use backstopjs. 

Advantages:
- you can add comments to config file
- blocks and be out-commented
- code can be generated dynamically
- more readable
- zillions of double quotes disappear in the config file

Disadvantages:
- ?

Here is a full example of "kind of" what I mean and how I would use it. 
https://github.com/klodoma/backstopjs-debug/tree/feature/js-config

![image](https://github.com/garris/BackstopJS/assets/944350/47f6140e-f628-4d09-bea2-59ce29238a3b)
![image](https://github.com/garris/BackstopJS/assets/944350/15c5fb3b-19d6-4c5f-9ccb-a8812fa9e882)

garris:
Yes. I would consider changing this convention.  It would require updating the docs and some of the scaffolding code.

My only concern would be that it adds some (probably minor) complexity for beginners or QA engineers with limited JS understanding.

maxfenton:
Could the default `backstop test` check for either of those files existing? As it stands I have to run `backstop test --config=backstop.config.js` every time. It seems like a check for `backstop.json` that falls back to `backstop.js` would be a nice improvement.

garris:
I believe it already does this.  If you run the sanity check it the docs it should work like that.

maxfenton:
@garris I think it does if it’s part of a build process but the CLI version does not. I might be wrong, and I'm not sure what I would need to do to change https://github.com/garris/BackstopJS/blob/f0651dc5439cebc1445a611115b7c16f4d2074a4/cli/index.js#L15 enough to make a PR.

My working process is to `npm install -g backstopjs` and then add a `backstop.js` configuration to each project directory and locally run `backstop` with that `--config` flag.

maxfenton:
Just wanted to follow up about the comment from August. BackstopJS does not have any kind of fallback system for config files names and only looks for `backstop.json` in [cli/index.js(https://github.com/garris/BackstopJS/tree/6d0e5fe42e15f6fe8468dced56087bf6d29cbbba/cli)

[...]

PR #1565, opened 2024-04-30
Use any config files: backstop.json, backstop.js

klodoma:
Use any config files: backstop.json, backstop.js whichever is found first. 

Discussion here: https://github.com/garris/BackstopJS/issues/1482

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -11,9 +11,7 @@ function main () {
   const argsOptions = parseArgs(process.argv.slice(2), {
     boolean: ['h', 'help', 'v', 'version', 'i', 'docker'],
     string: ['config'],
-    default: {
-      config: 'backstop.json'
-    }
+    default: {}
   });
 
   // Catch errors from failing promises
--- a/core/util/makeConfig.js
+++ b/core/util/makeConfig.js
@@ -15,8 +15,11 @@ function loadProjectConfig (command, options, config) {
   if (options.testReportFileName) {
     config.testReportFileName = options.testReportFileName;
   }
+
+  // list of "config" files to search for
   const configFiles = ['backstop.json', 'backstop.js'];
 
+  // If a config file is specified in the options, add it to the list
   if (options.config && typeof options.config === 'string') {
     configFiles.unshift(options.config);
     configFiles.unshift(path.join(config.projectPath, options.config));
--- a/core/util/makeConfig.js
+++ b/core/util/makeConfig.js
@@ -1,32 +1,38 @@
 const path = require('path');
+const fs = require('fs');
 const extendConfig = require('./extendConfig');
 
 const NON_CONFIG_COMMANDS = ['init', 'version', 'stop'];
 
-function projectPath (config) {
+function projectPath () {
   return process.cwd();
 }
 
 function loadProjectConfig (command, options, config) {
+  options = options || {}; // make sure options is an object
+
   // TEST REPORT FILE NAME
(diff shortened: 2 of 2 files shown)


core/util/makeConfig.js
```diff
@@ -1,32 +1,44 @@
...
-    config.testReportFileName = options.testReportFileName || null;
+  if (options.testReportFileName) {
+    config.testReportFileName = options.testReportFileName;
   }
 
-  let customConfigPath = options && (options.backstopConfigFilePath || options.configPath);
-  if (options && typeof options.config === 'string' && !customConfigPath) {
-    customConfigPath = options.config;
+  // list of "config" files to search for
+  const configFiles = ['backstop.json', 'backstop.js'];
```
maxfenton, line 21:
```suggestion
  const configFiles = ['backstop.json', 'backstop.config.json', 'backstop.js', 'backstop.config.js'];
```

maxfenton:
This code works for me locally and does exactly what I needed. Thank you.

core/util/makeConfig.js
klodoma: @maxfenton Do you REALLY need the support `backstop.config.*`? 

I personally don't mind if we add them, still I don't see it necessary. 

@garris  ? What do you say?

maxfenton: @klodoma I have a coworker who — today — said he prefers it on his projects because it’s more clear when looking at the root of a project what's a config file. I'd feel bad if I didn't at least make this suggestion for his sake and since it’s probably the default some other users have been using with the config flag.

maxfenton: And it’s the default name in the documentation here: https://github.com/garris/BackstopJS?tab=readme-ov-file#working-with-your-config-file

cli/index.js
```diff
@@ -11,9 +11,7 @@ function main () {
   const argsOptions = parseArgs(process.argv.slice(2), {
     boolean: ['h', 'help', 'v', 'version', 'i', 'docker'],
     string: ['config'],
-    default: {
-      config: 'backstop.json'
```
garris, line 15: @klodoma @maxfenton @dgrebb 

FWIW:  if we change this to `backstop` then require() will resolve this as a `.js` or a `json` file on its own.   This would be the absolute simplest change.  I don't think any other changes would be needed.

klodoma: @garris 

Please have a look at this PR. I'll add more comments in the PR. 
https://github.com/garris/BackstopJS/pull/1566

So probably we'll abandon this current PR.

maxfenton: I will test both today, but I think the docs on the new PR still suggest a `backstop.config.js` filename and I'm not clear if resolve will find a file with `.config` in it.

I do like that this PR makes it clear what filenames are allowed and what order of priority they would take.

maxfenton: I find this PR to be more elegant than v2 which adds a `useJs` flag that seems extraneous. I think the logic in this PR handles the case where someone has added both a `backstop.js` and `backstop.json` in the directory (the json is used because of array priority) and this PR handles more filenames and shows the order of them more explicitly.

That's just my opinion, but I'd vote for this PR over the other one, even though `resolve` can spot `.json` and `.js`

maxfenton: (And I really appreciate the work in both PR's @klodoma!)

klodoma:
I'll close this one. The V2 is the way to go.

Issue #1568, opened 2024-05-17
Values of requireSameDimensions and misMatchThreshold are ignored


abautu:
When I define any value (true or false) for requireSameDimensions, BackstopJS sees it as true (internally, it replaces that value with the default for misMatchThreshold, which is 0.1 and later coerced to true).
When I define any value (e.g. 0 or 0.2) for misMatchThreshold, BackstopJS ignores it and uses the default value which is 0.1.

Here is an snippet of the json file:
```
...
  "scenarioDefaults": {
      "requireSameDimensions": false,
      "misMatchThreshold": 0.2
  },
  "scenarios": [
    {
      "label": "About > Mission",
      "requireSameDimensions": false,
      "misMatchThreshold": 0.2,
....
```

PR #1569, opened 2024-05-17
#1568 Fix defaults in getRequireSameDimensions

abautu:
Make getRequireSameDimensions return either defaultRequireSameDimensions or true (instead of 0.1)


garris (approved):
Looks like a good fix!

PR #1614, opened 2025-11-25
chore: Updates to stable Debian and node 24 LTS

mradcliffe:
I was trying to install some new packages into the docker container on bullseye, but I found bullseye and even bookworm were too old.

This PR updates:

- node LTS release (24)
- Debian stable (trixie)

I ran the following locally:

- npm run init-docker-builder
- npm run build-docker
- npm run build-and-load-docker
- npm run sanity-test-docker
- npm run smoke-test-docker

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # use bullseye node base, as debian does provide a chromium for arm64 and amd64 flavour
-FROM node:20-bullseye
+FROM node:24-trixie
 
 ARG BACKSTOPJS_VERSION
 ARG DEBIAN_FRONTEND=noninteractive


garris (changes requested):
Is it possible to use Debian 12 stable?

mradcliffe:
> Is it possible to use Debian 12 stable?

I did some more investigation, and it seems like I can manage with a third-party repository to get the packages I need on Debian 12. Thank you.

Since I made my branch name Debian 13 specific, should I close this and then re-open a merge request with a better named branch?

garris:
Something like chore/upgrade-base-image would be clearer.  Thank you.

mradcliffe:
Replaced by #1616 per garris review and suggestion.

